PR #65, opened 2026-07-08
Port upstream list paste ordering fix

Renakoni:
## Summary
- Port upstream MarkText PR #4842 into the Android Muya copy.
- Keep pasted list items in order when merging a pasted list into the middle of another list.
- Add regression coverage for bullet and ordered list middle-paste cases.

## Upstream Source
- Upstream PR: https://github.com/marktext/marktext/pull/4842
- Upstream issue: https://github.com/marktext/marktext/issues/3549

## Android Relevance
This affects Android's live Muya editor paste behavior. Before this fix, `tryMergeListPaste()` rebuilt the enclosing list by appending pasted items to the end of the list, even when the cursor was in a middle item. Pasting `- Item 1 / - Item 2 / - Item 3` into an empty middle item between `Item A` and `Item B` could produce `A, Item 1, B, Item 2, Item 3` instead of keeping the pasted sequence together.

## Change Details
- Splice pasted list items immediately after the anchor item instead of pushing them to the end.
- Extract post-rebuild caret placement into `seatListMergeCursor()` so the caret lands on the folded paragraph, trailing pasted states, or the last pasted list item as appropriate.
- Add tests for bullet-list and ordered-list middle paste ordering.
- Android currently preserves source ordered-list markers during serialization, so the ordered-list regression locks item order while preserving the repo's existing numbering policy.

## Verification
- `pnpm --dir third_party/muya exec vitest run src/clipboard/__tests__/pasteListMerge.spec.ts` passed.
- `pnpm build` passed.
- Broader clipboard suite command `pnpm --dir third_party/muya exec vitest run src/clipboard/__tests__` was also run; it has one pre-existing unrelated failure in `cutSelectionToClipboardData.spec.ts` where `cutSelectionToClipboardData` does not delete `alpha` from `alpha beta`. The touched paste-list spec passes independently.

### DIFF
--- a/third_party/muya/src/clipboard/__tests__/pasteListMerge.spec.ts
+++ b/third_party/muya/src/clipboard/__tests__/pasteListMerge.spec.ts
@@ -133,4 +133,24 @@ describe('paste — same-type list merges into the enclosing list (A5, muyajs pa
         expect(selection.anchor?.offset).toBe(4);
         expect(selection.focus?.offset).toBe(4);
     });
+
+    // #3549: pasting into a non-last item must keep the pasted items in order
+    // right after the anchor item, not append them to the end of the list.
+    it('inserts pasted items after the anchor item, not at the list end', async () => {
+        const muya = bootMuya('- Item A\n- \n- Item B\n');
+        const empty = contentBlocks(muya).find(b => b.text === '')!;
+        expect(await paste(muya, empty, 0, 0, '- Item 1\n- Item 2\n- Item 3')).toBe(
+            '- Item A\n- Item 1\n- Item 2\n- Item 3\n- Item B\n',
+        );
+    });
+
+    it('keeps order for an ordered list pasted into the middle', async () => {
+        const muya = bootMuya('1. A\n2. \n3. B\n');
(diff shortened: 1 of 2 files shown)


Renakoni:
Reviewed for the Android client: this ports upstream #4842 into the vendored Muya paste path, which is live editor behavior on Android (not desktop-only). The core fix (splice pasted items after the anchor instead of appending) is correct, and the extracted `seatListMergeCursor()` also covers the previously undefined caret case where a multi-item list is pasted with no trailing states — index math checks out for both the folded and non-folded paths.

One issue needed fixing: the branch was based on a main revision that predates the `test:app-gate` script, so the Muya CI gate could not even run against it. Rebased onto current main (clean, no conflicts). Verified after rebase:

- `pasteListMerge.spec.ts` targeted run: 8/8 (the two new middle-paste regressions included)
- `pnpm test:muya` gate: 201 files / 1390 tests green; the new spec is inside the gate scope, and the `cutSelectionToClipboardData` failure mentioned in the description is confirmed pre-existing and explicitly excluded from the gate
- App-side `pnpm test` (258) and production build green with the Muya change synced

Ready to merge once CI re-runs on the rebased head.

PR #67, opened 2026-07-08
Port upstream table escaped pipe fix

Renakoni:
## Summary
- Port upstream MarkText PR #4850 into the Android Muya copy.
- Fix escaped pipes inside table-cell inline code so the editor displays `|` / `||` instead of `\|` / `\|\|`.
- Add regression coverage for editor rendering, table structure, and markdown round-trip behavior.

## Upstream Source
- Upstream PR: https://github.com/marktext/marktext/pull/4850
- Upstream issue: https://github.com/marktext/marktext/issues/4849

## Android Relevance
This directly affects Android's live Muya editor rendering. GFM table cells escape literal pipes as `\|`; after parsing, that escape should be stored as a literal pipe. The previous import path re-added `\|` into the stored table-cell text. That looked acceptable in plain cell text because the inline escape rule hides the backslash, but inside inline code the backslash is literal, so Android users saw code cells like `\|` instead of `|`.

## Change Details
- Remove table-cell import-time re-escaping from `MarkdownToState`.
- Store table-cell text as marked emits it, with `\|` already resolved to literal `|`.
- Keep round-trip behavior through the existing `stateToMarkdown.escapeText` serialization path, which re-escapes pipes when writing markdown.
- Add `tableEscapedPipe.spec.ts` covering:
  - editor code rendering for `` `\|` `` and `` `\|\|` ``;
  - table row/cell structure preservation;
  - serialization round-trip for inline-code and plain-text escaped pipes.

## Verification
- `pnpm --dir third_party/muya exec vitest run src/state/__tests__/tableEscapedPipe.spec.ts` passed.
- `pnpm build` passed.

### DIFF
--- a/third_party/muya/src/state/__tests__/tableEscapedPipe.spec.ts
+++ b/third_party/muya/src/state/__tests__/tableEscapedPipe.spec.ts
@@ -1,0 +1,69 @@
+// @vitest-environment happy-dom
+
+import { describe, expect, it, vi } from 'vitest';
+import { Muya } from '../../muya';
+import { MarkdownToState } from '../markdownToState';
+import ExportMarkdown from '../stateToMarkdown';
+
+vi.mock('../../utils/prism/index', () => ({
+    default: {},
+    walkTokens: () => null,
+    loadedLanguages: new Set(),
+    transformAliasToOrigin: (language: string) => language,
+    loadLanguage: () => Promise.resolve([]),
+    search: () => [],
+}));
+
+function boot(markdown: string): Muya {
+    const host = document.createElement('div');
+    document.body.appendChild(host);
+    const muya = new Muya(host, { markdown } as ConstructorParameters<typeof Muya>[1]);
+    muya.init();
+    return muya;
+}
+
+function codeTexts(muya: Muya): string[] {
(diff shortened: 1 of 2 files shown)


Renakoni:
Reviewed for the Android client: relevant — this is live Muya table rendering, not desktop-only. The port matches upstream #4850: marked already resolves `\|` to a literal pipe during table tokenization, so the import-time `restoreTableEscapeCharacters()` re-escape was double-handling it, which showed up as literal `\|` inside table-cell inline code on Android. Removing it is correct, and I verified the round-trip guarantee it relies on is real: `stateToMarkdown.escapeText` re-escapes unescaped pipes on every table cell at serialization (with a lookbehind that prevents double-escaping), and the new spec locks editor rendering, table structure, and both round-trip cases.

Branch maintenance: the branch carried a stale merge-of-main (pre-#83) with an inconsistent committer identity, so I rebased onto current main instead — history is now the single port commit, and I confirmed the effective diff is byte-identical before/after the rebase. Verified on the new base: targeted spec 4/4, Muya gate 202 files / 1394 tests green (the new spec is in gate scope), app unit tests 258 and production build green.

Ready to merge once CI re-runs.

PR #69, opened 2026-07-08
Port upstream Prism dependency order fix

Renakoni:
## Upstream

Ports marktext/marktext#4861: https://github.com/marktext/marktext/pull/4861

Upstream fixed Muya's Prism language loader so dependency languages are loaded in the order Prism expects. The concrete failure is cold-loading C++ before C: `prism-cpp` extends the `c` grammar, so a parallel/racy load can fail before the dependency exists.

## Android relevance

Android uses this Muya Prism path for editor code highlighting in the WebView. The same dependency-order race applies here on cold grammar loads, especially C/C++ and other grammars with Prism dependencies.

## Changes

- Replace the deferred-status collection with an ordered async `loadComponent` callback.
- Pass Prism's dependency loader explicit `series` and `parallel` chainers so dependent components are awaited before dependents run.
- Keep returned load statuses for `loaded`, `cached`, and `noexist` outcomes.
- Add a focused regression test that loads `cpp`, asserts `c` is registered before `cpp`, and verifies tokenization does not throw.
- Android-specific correction: resolve Prism component dynamic imports from Muya's package location with `../../../node_modules/prismjs/components/prism-*.js`. The previous glob made the focused Vitest run unable to resolve several Prism modules such as `c`, `latex`, and `yaml`; this path keeps the dynamic import map tied to the actual dependency under this workspace.

## Verification

- `pnpm --dir third_party/muya exec vitest run src/utils/prism/__tests__/loadLanguageDependencyOrder.spec.ts`
- `pnpm build`

### DIFF
--- a/third_party/muya/src/utils/prism/__tests__/loadLanguageDependencyOrder.spec.ts
+++ b/third_party/muya/src/utils/prism/__tests__/loadLanguageDependencyOrder.spec.ts
@@ -1,0 +1,37 @@
+// @vitest-environment happy-dom
+
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import prism, { loadLanguage } from '../index';
+import { loadedLanguages } from '../loadLanguage';
+
+function resetCppState() {
+    for (const lang of ['c', 'cpp']) {
+        delete (prism.languages as Record<string, unknown>)[lang];
+        loadedLanguages.delete(lang);
+    }
+}
+
+afterEach(() => {
+    vi.restoreAllMocks();
+});
+
+describe('loadLanguage dependency order', () => {
+    it('loads a dependency before the dependent grammar that extends it', async () => {
+        resetCppState();
+
+        const addOrder: string[] = [];
+        const realAdd = loadedLanguages.add.bind(loadedLanguages);
+        vi.spyOn(loadedLanguages, 'add').mockImplementation((lang: string) => {
+            addOrder.push(lang);
+            return realAdd(lang);
+        });
+
+        await expect(loadLanguage('cpp')).resolves.toBeDefined();
+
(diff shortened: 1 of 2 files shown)


Renakoni:
Reviewed for the Android client: the ordering fix itself is a faithful port of upstream #4861 and is Android-relevant (WebView code highlighting). Passing an async `loadComponent` with explicit `series`/`parallel` chainers is the correct way to make Prism's dependency resolver await grammar imports, and the cpp-extends-c regression test locks it.

**Found and fixed a runtime regression in the "Android-specific correction":** switching the module glob to `../../../node_modules/prismjs/components/...` fixed Muya's local vitest context but silently emptied the module map in the Android app build — `@muyajs/core` is a scoped package compiled from the pnpm store copy, where prismjs resolves five levels up beside the package, not three levels into it. `pnpm build` cannot catch this (an empty glob map is not a build error); the fenced-code E2E does: with this branch as pushed, `renders loaded math, fenced code, and Mermaid previews` failed because the TypeScript grammar never loaded, i.e. every cold grammar load on Android would have thrown `Prism language module not found`.

Fix (amended into the port commit): keep both glob shapes as **two separate `import.meta.glob` calls merged into one map** — the original five-up shape for the app/pnpm-store context and the new three-up shape for Muya's own context. They must stay separate calls: a control test showed that mixing the out-of-root shape into a single glob array poisons resolution for the Muya context too. The existing `getLanguageModule` endsWith fallback already handles both key shapes.

Also rebased onto current main (branch was behind). Verified after the fix: dependency-order spec 1/1, Muya gate 203 files / 1395 green, full markdown-rendering E2E 5/5 (proves app-context grammar loading), app unit 258, production build green.

PR #70, opened 2026-07-08
Port upstream code fence info string model

Renakoni:
## Upstream

Ports marktext/marktext#4856: https://github.com/marktext/marktext/pull/4856

This is stacked on #69 (`sync/upstream-4861-prism-dependency-order`) because the new live Muya tests import the editor Prism path, and Android main still has the pre-#4861 Prism dynamic import issue that causes unrelated `latex` / `yaml` module resolution errors under Vitest.

## Android relevance

Android uses Muya for the WebView editor, including fenced code block parsing, live language input editing, serialization, and Prism highlighting. The upstream bug affects mobile editor behavior directly:

- ` ```js title="app.js" ` is currently reduced to `js` in the model.
- Editing the language input truncates at the first whitespace and drops attributes.
- Using the whole info string as a CSS `language-*` token can crash or produce invalid class handling unless highlighting derives only the first word.

## Changes

- Treat `meta.lang` as the full fenced code info string, matching CommonMark's model.
- Add `firstWordOfInfo()` and use it where the actual syntax-highlight language is needed.
- Load Prism from the derived first word, not from the full info string.
- Render/highlight code blocks using the derived language token so attributes do not leak into `language-*` classes.
- Let the language input edit and persist the whole info string instead of truncating it.
- Keep markdown serialization emitting `meta.lang` verbatim.
- Add focused regression coverage for parser model, helper behavior, live code-block rendering, and language-input editing.

## Verification

- `pnpm --dir third_party/muya exec vitest run src/utils/__tests__/firstWordOfInfo.spec.ts src/state/__tests__/infoStringModel.spec.ts src/block/content/codeBlockContent/__tests__/infoStringHighlight.spec.ts src/block/content/langInputContent/__tests__/editInfoString.spec.ts src/state/__tests__/blockSerialization.spec.ts`
- `pnpm build`

### DIFF
--- a/third_party/muya/src/block/commonMark/codeBlock/index.ts
+++ b/third_party/muya/src/block/commonMark/codeBlock/index.ts
@@ -2,7 +2,7 @@ import type { Muya } from '../../../muya';
 import type { ICodeBlockState } from '../../../state/types';
 import type { TBlockPath } from '../../types';
 import diff from 'fast-diff';
-import { diffToTextOp } from '../../../utils';
+import { diffToTextOp, firstWordOfInfo } from '../../../utils';
 import { operateClassName } from '../../../utils/dom';
 import logger from '../../../utils/logger';
 import { loadLanguage } from '../../../utils/prism';
@@ -72,8 +72,10 @@ class CodeBlock extends Parent {
             operateClassName(this.domNode!, 'add', 'mu-fenced-code');
         }
 
-        !!value
-        && loadLanguage(value)
+        // `value` is the full info string; load Prism for its first word only.
+        const language = firstWordOfInfo(value);
+        !!language
+        && loadLanguage(language)
             .then((infoList) => {
                 if (!Array.isArray(infoList))
                     return;
--- a/third_party/muya/src/block/content/codeBlockContent/__tests__/infoStringHighlight.spec.ts
+++ b/third_party/muya/src/block/content/codeBlockContent/__tests__/infoStringHighlight.spec.ts
@@ -1,0 +1,31 @@
+// @vitest-environment happy-dom
+import { describe, expect, it } from 'vitest';
+import { Muya } from '../../../../muya';
+
+function boot(markdown: string): Muya {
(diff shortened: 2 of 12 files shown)


Renakoni:
Reviewed for the Android client: relevant (live fenced-code parsing, language-input editing, highlighting) and the port itself is faithful to upstream #4856 — `meta.lang` becomes the verbatim info string, `firstWordOfInfo()` derives the highlight language at the Prism-load and `language-*` class sites, and the lang input stops truncating attributes on edit.

Two problems needed fixing before this could land:

1. **The branch was stacked on the pre-fix #69 commit.** Its history carried the original `sync/upstream-4861` commit whose glob change broke app-context Prism loading (fixed before #69 merged). Rebased with `--onto` so only the info-string commit replays on current main — a plain rebase would have resurrected the broken glob.
2. **Head-on collision with the already-merged #68 (upstream #4770).** Main had the interim model (`lang` = first word + separate `meta.info` with a staleness guard); this PR implements the upstream model (`lang` = whole info string). The rebase applied "cleanly" but produced a duplicate `const info` declaration (compile error) and left both mechanisms half-alive. Reconciled toward the upstream model, which is the right long-term shape for a fork that keeps porting upstream code-block changes: removed `meta.info`, its serializer guard, and its type field; kept #4770's four round-trip regressions (they now validate the new model — coverage continuity); dropped only the stale-info guard test, whose failure mode is impossible once there is a single source of truth.

Also audited every remaining `meta.lang` consumer under the full-string model: the OT language-input sync stores the verbatim string (intended), highlighting and Prism loading go through `firstWordOfInfo`/alias resolution before touching `language-*` classes, diagram detection matches on the first word, and the frontmatter serializer switch operates on frontmatter langs, not fence info strings.

Verified: info-string spec battery 41/41 (including the updated #4770 spec), Muya gate 207 files / 1406 green, markdown rendering + editing E2E 9/9 (fenced code, math, Mermaid detection under the new model), app unit 258, production build green.

PR #71, opened 2026-07-08
Port upstream heading insert menu labels

Renakoni:
## Upstream

Ports marktext/marktext#4854: https://github.com/marktext/marktext/pull/4854

This is stacked on #69 (`sync/upstream-4861-prism-dependency-order`) only because the focused Muya test boots editor code that imports the Prism path, and Android main still has the pre-#4861 Prism dynamic import issue under Vitest.

## Android relevance

Android carries Muya's paragraph quick insert menu and locale files. The menu is editor UI, so the terminology fix applies to the WebView editor as well: h1-h6 are CommonMark/HTML headings, not page headers.

## Changes

- Rename the quick insert menu section from `headers` to `headings`.
- Rename display/i18n keys from `Header 1` ... `Header 6` to `Heading 1` ... `Heading 6`.
- Update all Muya locale key names while preserving existing translated values.
- Extend quick insert menu search coverage to assert the new section name and heading titles.

## Verification

- `pnpm --dir third_party/muya exec vitest run src/ui/paragraphQuickInsertMenu/__tests__/search.spec.ts`
- `pnpm build`


Renakoni:
Reviewed for the Android client: relevant — the paragraph quick insert menu is an active Android surface (the app exposes a user-facing `quickInsert` Editing setting, on by default, and the menu renders inside the WebView editor), so the Header→Heading terminology fix from upstream #4854 applies here, zh-CN locale included.

Rename completeness verified: zero leftovers of `Header 1..6` / `headers` anywhere in Muya source; all ten locale files rename the lookup keys while preserving the translated values; the functional `atx-heading N` replacement labels are untouched, so this is purely display-layer.

Branch maintenance: like the other stacked ports, this branch sat on the pre-fix #69 commit whose Prism glob change was corrected before merging — rebased with `--onto` so only the label commit replays on current main. Verified after rebase: quick-insert search + replace-by-label specs 23/23, Muya gate 207 files / 1406 green, app unit 258, production build green.

Ready to merge once CI re-runs.

PR #73, opened 2026-07-08
Port upstream linked image modifier click fix

Renakoni:
## Upstream

Ports marktext/marktext#4858: https://github.com/marktext/marktext/pull/4858

This is stacked on #69 (`sync/upstream-4861-prism-dependency-order`) because the focused Muya tests boot the editor and Android main still has the pre-#4861 Prism dynamic import issue under Vitest.

## Android relevance

Android carries the same Muya link and image click handlers. Modifier-click is primarily desktop-style, but Android devices can use hardware keyboards, mice, tablets, and Chromebooks; keeping Muya's event contract aligned also prevents the image preview handler from racing the link handler for linked images.

## Changes

- Export Muya's `LINK_SELECTOR` so image selection can reuse the same definition of link wrappers.
- Suppress image-preview `format-click` emission when a modifier-clicked image is inside a link wrapper.
- Preserve existing plain-image modifier-click preview behavior.
- Add regression coverage for linked images and plain images.

## Verification

- `pnpm --dir third_party/muya exec vitest run src/selection/__tests__/linkedImageClick.spec.ts src/selection/__tests__/duplicateImageResize.spec.ts src/selection/__tests__/parityPreviewImage.spec.ts src/__tests__/formatClickEvents.spec.ts`
- `pnpm build`

### DIFF
--- a/third_party/muya/src/editor/linkMouseEvents.ts
+++ b/third_party/muya/src/editor/linkMouseEvents.ts
@@ -36,7 +36,7 @@ import { getLinkInfo } from '../utils/getLinkInfo';
 // `mu-raw-html` is added to every inline HTML tag (`<u>`, `<mark>`,
 // `<sub>`, `<sup>`, `<a>` …), so we can't match it loosely — narrow
 // each entry by the actual tag the renderer emits.
-const LINK_SELECTOR = [
+export const LINK_SELECTOR = [
     `span.${CLASS_NAMES.MU_LINK}`,
     `a.${CLASS_NAMES.MU_REFERENCE_LINK}`,
     `a.${CLASS_NAMES.MU_RAW_HTML}`,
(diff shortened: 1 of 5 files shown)


Renakoni:
Reviewed for the Android client: recommending closure as a desktop-only interaction.

The entire behavioral change sits inside the `event.metaKey || event.ctrlKey` branch of `ImageSelection` — without a physical modifier key the code paths are byte-identical before and after this PR, so the fix is unreachable for the touch-based usage this client targets. For the residual hardware-keyboard/mouse case (tablets, Chromebooks, DeX), the Android app has no subscriber for the affected editor events (`src/` contains no `format-click`/link-open consumers), so both the suppressed preview emission and the link navigation it raced are no-ops here — upstream #4858 fixes a conflict between two desktop behaviors, neither of which is wired up on Android.

Note for the record: #74 stacks on this branch but is a different animal — it fixes illegal `<a>`-in-`<a>` nesting that structurally breaks reference-linked image rendering, which is Android-relevant regardless of input device. It will be rebased off this branch (and off the pre-fix #69 commit it also carries) when it is processed; its only ties to this PR are additions to the spec file this branch introduced, which can be carried standalone.

If a link-opening/image-preview feature ever lands on Android, this port can be revived from upstream #4858.

PR #74, opened 2026-07-08
Port upstream reference linked image anchor fix

Renakoni:
## Upstream

Ports marktext/marktext#4866: https://github.com/marktext/marktext/pull/4866

This is stacked on #73 (`sync/upstream-4858-linked-image-modifier-click`) because upstream #4866 depends on #4858's exported `LINK_SELECTOR`, linked-image modifier-click guard, and linked-image regression test.

## Android relevance

Android carries the same Muya inline tokenizer, image renderer, CSS, and image/link click handlers. Reference-linked README badges (`[alt][ref]`) and raw-HTML linked images can appear in Markdown edited on Android, so the same rendering and event behavior applies in the WebView editor.

## Changes

- Allow `reference_link` tokenization to capture image link text, so `[alt][ref]` becomes one reference link wrapping the image instead of a bare image plus fragmented reference link.
- Render inline-image hover controls as `span` instead of nested `a` elements, preventing invalid `<a>` inside `<a>` markup when an image is inside a real reference/raw-HTML anchor.
- Relax the image icon CSS selectors from `a.mu-image-icon-*` to `.mu-image-icon-*` to match the new span controls.
- Extend linked image click coverage for reference-linked images.
- Add tokenizer coverage for image reference-link anchors.
- Add jsdom coverage for raw-HTML linked images, where the real anchor nesting behavior is observable.

## Verification

- `pnpm --dir third_party/muya exec vitest run src/inlineRenderer/__tests__/referenceLinkImageAnchor.spec.ts src/selection/__tests__/linkedImageClick.spec.ts src/selection/__tests__/rawHtmlLinkedImageClick.spec.ts`
- `pnpm build`

### DIFF
--- a/third_party/muya/src/assets/styles/inlineSyntax.css
+++ b/third_party/muya/src/assets/styles/inlineSyntax.css
@@ -408,9 +408,9 @@ blockquote .mu-hide.mu-math > .mu-math-render {
     background: transparent;
 }
 
-.mu-inline-image a.mu-image-icon-success,
-.mu-inline-image a.mu-image-icon-fail,
-.mu-inline-image a.mu-image-icon-close {
+.mu-inline-image .mu-image-icon-success,
+.mu-inline-image .mu-image-icon-fail,
+.mu-inline-image .mu-image-icon-close {
     position: absolute;
     top: 15px;
 
@@ -419,8 +419,8 @@ blockquote .mu-hide.mu-math > .mu-math-render {
     height: 20px;
 }
 
-.mu-inline-image a.mu-image-icon-success,
-.mu-inline-image a.mu-image-icon-fail {
+.mu-inline-image .mu-image-icon-success,
+.mu-inline-image .mu-image-icon-fail {
     left: 15px;
 }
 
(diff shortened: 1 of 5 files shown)


Renakoni:
Reviewed for the Android client — **recommending merge, not closure**, despite the "click" in the original title. The click framing came from the upstream PR chain; the substance here is device-independent rendering correctness:

1. **Tokenizer**: `reference_link` could not hold brackets in its link text, so the ubiquitous README badge form `[alt][ref]` fragmented into a bare image plus broken reference link instead of one link wrapping the image.
2. **Renderer**: the image hover controls rendered as `<a>` elements; inside a real anchor (reference link or raw HTML) the illegal `<a>`-in-`<a>` nesting makes the HTML parser close the outer anchor early, structurally hoisting the image out of its link in the live WebView DOM.

Proof the bug is real on current main: running this PR's regression specs against main's sources fails 4 of 6 (badge syntax fragments, both raw-HTML and reference-linked images escape their anchors, `<a>` controls present); with the fix, 6/6 pass.

Surgery performed before push:
- Rebased with `--onto`, dropping the two stacked commits this branch carried (the pre-fix #69 commit and closed #73).
- The upstream chain's modifier-click routing is **not** ported, consistent with closing #73: the click-suppression assertions and the dependency on #73's spec file are gone. The regression coverage is rewritten as pure structure specs (`linkedImageAnchor.spec.ts`, renamed from `rawHtmlLinkedImageClick.spec.ts`) — image stays inside `a.mu-raw-html` / `a.mu-reference-link`, and no anchor-tagged hover controls exist — plus the untouched tokenizer spec.
- Commit message and PR title updated to describe what actually lands.

Verified: anchor + tokenizer specs 6/6, Muya gate 209 files / 1412 green, markdown-rendering E2E 5/5, app unit 258, production build green.

PR #87, opened 2026-07-10
Extract home document actions from App.vue

Renakoni:
## Summary

Second PR of the semantic-decomposition series: moves the home selection-mode action orchestration (pin, delete, share, rename) out of `App.vue` into a `createHomeDocumentActions` factory in `features/home`. Behavior-preserving — no product behavior, storage format, notice copy, or platform payload changes.

## Boundary

The factory owns exactly the four selection actions and their two private helpers (selected-record lookup, stored-id collection for pin pruning):

- **Pin**: Telegram-style toggle over the selection via the existing `togglePinnedDocuments`, stale-notice clearing, selection exit.
- **Delete**: drafts removed permanently, device files removed from Recents only, pins pruned against everything in storage (hidden recovery/setting-disabled drafts keep theirs), open editor session reset when its document is deleted.
- **Share**: existing `shareRecentDocumentsWorkflow` unchanged; failure keeps the selection for retry, success clears it.
- **Rename**: draft rename with open-session displayName patching; Android rename through the existing workflow with URI migration of the recent record, pin, recovery draft, and open session, including the temporary-access drop path and its notice.

State stays in App and is injected as refs (stores + persistence, selection instance, document state, notice); native adapters and settings are injected the same way the workflows already consume them. Pure helpers (`togglePinnedDocuments`, `renameLocalDraft`, `getAndroidRecoveryDraftId`, …) are imported directly by the module. The template bindings are untouched — App destructures the same four handler names. The one inlining: the mount-time pin prune keeps a local two-line stored-id collection with its explanatory comment.

`App.vue` shrinks by ~190 lines.

## Testing

- New `homeDocumentActions.test.ts` (10 tests) drives the real share/rename workflows with mocked natives: pin toggle/empty-selection no-op, permanent draft delete vs list-only device delete with pin pruning, open-session reset on delete, share failure keeping the selection vs success clearing it, draft rename patching the open session, URI-changing rename migrating record/pin/recovery draft, temporary-access rename dropping entry and pins with its notice, and missing-target warn.
- Full unit suite 276/276, ESLint, vue-tsc, production build green.
- Contract E2E: `mobile-home-document-management.spec.ts` (long-press selection lifecycle with delete, temporary-access rename storage assertions, recovery-draft survival) plus `mobile-home-navigation.spec.ts` — 5/5.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -115,7 +115,6 @@ import {
 } from './features/editor/editorRuntime'
 import {
   removeLocalDraft,
-  renameLocalDraft,
   upsertLocalDraft,
   type LocalDraftRecord,
 } from './lib/localDrafts'
@@ -133,7 +132,6 @@ import {
   areAllDocumentsPinned,
   getPinnedDocumentIds,
   prunePinnedDocuments,
-  togglePinnedDocuments,
   type PinnedDocumentRecord,
 } from './lib/pinnedDocuments'
 import {
@@ -154,8 +152,7 @@ import {
   type HomeDocumentText,
 } from './features/home/homeDocuments'
 import { useDocumentSelection } from './features/home/useDocumentSelection'
-import { renameAndroidRecentDocumentWorkflow } from './features/android-documents/renameRecentDocumentWorkflow'
-import { shareRecentDocumentsWorkflow } from './features/android-documents/shareRecentDocumentsWorkflow'
+import { createHomeDocumentActions } from './features/home/homeDocumentActions'
 import {
   DEFAULT_SETTINGS_PAGE,
   SETTINGS_PAGES,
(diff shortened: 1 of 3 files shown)


Renakoni:
Review finding confirmed and fixed: the extracted `pinSelectedDocuments` and `deleteSelectedDocuments` had picked up a `homeNotice.value = null` that the pre-refactor App.vue never performed — an undocumented notice-lifecycle change smuggled into a parity refactor, and the new tests had locked it in as expected. Both clears are removed; the tests now initialize a non-empty notice and assert it survives pin and delete untouched. The share and rename notice handling is unchanged — those paths cleared/updated the notice before the refactor, so their current form is a faithful migration.

Re-verified after the fix: module tests 10/10 (with the corrected notice-preservation assertions), full unit suite 276/276, ESLint, vue-tsc, production build, and the home-document-management + home-navigation E2E suites 5/5.

PR #94, opened 2026-07-10
Resolve derived titles to plain text

Renakoni:
## Summary

Fixes the clunky titles reported for styled first lines: a draft starting with `***<u>alpha</u>*** bravo charlie delta` used to lose its emphasis markers but keep the raw `<u>` tags, so the home list showed `<u>alpha</u> bravo charlie delta`. Derived titles now always resolve to plain text — `<u>alpha</u> …` and `***<u>alpha</u>*** …` both title as `alpha bravo charlie delta`.

Kept deliberately separate from #93: this is a user-visible behavior fix on the web side, while the decomposition series PRs are behavior-preserving by contract.

## Change

- New shared `stripInlineMarkdown` reduction in `documentState.ts`: keeps link/image text, drops inline HTML tags and emphasis markers, collapses whitespace.
- Applied on **both** title paths, since the same leak existed one `#` away: the leading-line heuristic from #83 and the heading branch (`# **Trip** to <u>Paris</u>` now titles as `Trip to Paris`).
- The tag pattern matches raw HTML the way CommonMark does (letter-led tag name, optional attributes, optional self-close), so autolinks like `<https://example.com>` are treated as content and preserved.
- A heading that reduces to pure markup (`# ***`) now names nothing and falls through to the display-name/leading-text fallbacks instead of titling the document as a row of asterisks.

Titles are derived, never stored as user input, so existing documents pick up clean titles on their next open/save with no migration.

## Testing

- Three new/extended unit cases lock the reported case verbatim (with and without emphasis), tag variants (`<mark>`, `<br>` word spacing), autolink preservation, styled-heading reduction, and the pure-markup-heading fall-through. `documentState` suite 23/23.
- Full unit suite 310/310, ESLint, vue-tsc, production build green.

### DIFF
--- a/src/lib/documentState.test.ts
+++ b/src/lib/documentState.test.ts
@@ -34,6 +34,43 @@ describe('documentState', () => {
     expect(getDocumentTitle('1. step one\n2. step two')).toBe('step one')
   })
 
+  it('drops inline HTML tags from derived titles', () => {
+    // Styled first lines always resolve to plain text — with or without
+    // emphasis around the tags, the title is the same readable content.
+    expect(getDocumentTitle('***<u>alpha</u>*** bravo charlie delta'))
+      .toBe('alpha bravo charlie delta')
+    expect(getDocumentTitle('<u>alpha</u> bravo charlie delta'))
+      .toBe('alpha bravo charlie delta')
+    expect(getDocumentTitle('<mark>note</mark> for later')).toBe('note for later')
+    expect(getDocumentTitle('alpha<br>beta')).toBe('alpha beta')
+    // Autolinks are content, not markup.
+    expect(getDocumentTitle('<https://example.com> reading list'))
+      .toBe('<https://example.com> reading list')
+  })
+
+  it('reduces styled headings to plain-text titles', () => {
+    expect(getDocumentTitle('# **Trip** to <u>Paris</u>')).toBe('Trip to Paris')
+    // A pure-markup heading names nothing; the content below wins.
+    expect(getDocumentTitle('# ***\n\nreal text')).toBe('real text')
+  })
+
+  it('keeps literal Markdown punctuation that is not paired markup', () => {
+    expect(getDocumentTitle('# file_name')).toBe('file_name')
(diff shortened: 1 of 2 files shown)


Renakoni:
Both review points addressed in the updated commit (also rebased onto current main, so the merge commit is gone and the branch is a single commit again):

**1. Literal punctuation preserved.** The global `[`*_~]` deletion is gone. `stripInlineMarkdown` now only unwraps actual paired delimiters:
- code spans `` `...` ``
- asterisk emphasis `*{1,3}...*{1,3}`
- underscore emphasis with CommonMark's intraword rule — `_` flanked by letters/digits (`(?<![\p{L}\p{N}])_{1,3}([^_]+)_{1,3}(?![\p{L}\p{N}])`) never counts as markup
- GFM strikethrough `~~...~~` only (single `~` is content)

So `# file_name` → `file_name`, `# 2*3 benchmark` → `2*3 benchmark`, `# foo~bar` → `foo~bar`, and suggested filenames keep their underscores. One behavior note: with paired unwrapping, a marker-only heading like `# ***` no longer collapses to empty by accident, so the helper now explicitly treats a reduction consisting solely of marker characters as "no title" — the existing `# ***` → fall-through-to-content test kept passing because of this rule.

**2. Inline HTML tag pattern fixed.** Now `/<\/?[a-z][a-z0-9-]*(?:[\s\/](?:[^<>"']|"[^"]*"|'[^']*')*)?>/gi`:
- hyphenated custom elements match (`<my-widget>alpha</my-widget> beta` → `alpha beta`)
- quoted attribute values may contain `>` (`<span title="a>b">alpha</span> beta` → `alpha beta`, no leaked `b">`)
- autolinks still don't match because the character after the tag name must be whitespace, `/`, or `>` — `<https://…>` and `<user@…>` remain content (covered by tests)

Added the three suggested heading regressions plus `snake_case_name`, `~~done~~`, `_emphasized_`, the custom-element case, the quoted-attribute case, and an email-autolink case. Full suite 312/312, lint and build clean.

PR #98, opened 2026-07-11
Add mobile in-document search to the editor

Renakoni:
## Summary

Implements the editor's Search action, which was a logging placeholder (`openEditorSearch()` only wrote `editor search requested`). Tapping the top-bar search icon now swaps the top bar for a touch-friendly find bar: query input (auto-focused), a live `n/m` match counter, previous/next navigation with wrap-around, and a close button.

The engine is Muya's existing search module, driven through its public API — no Muya internals were modified:

- `muya.search(value)` collects matches and renders highlights (`span.mu-highlight` for the active match, `span.mu-selection` for the rest; styles ship with Muya's own CSS).
- `muya.find('previous' | 'next')` navigates with wrap-around; the app scrolls the active highlight into view (Muya does not scroll on its own).
- Closing runs `muya.search('', { selectHighlight: true })` — Muya's documented contract that clears every highlight and drops the editor cursor onto the last active match, so the editing position is preserved (covered upstream by `search.spec.ts`).

App-side, a new `features/editor/documentSearch.ts` session controller owns the open/query/navigate/close state (injectable `getEditor` + scroll callback, mirroring the existing extraction pattern). The typed `@muyajs/core` contract in `muya-core.d.ts` gains `search`/`find` and the option/state interfaces.

## Lifecycle and Back behavior

- Android Back closes the search bar (new `close-editor-search` decision in `getAppBackButtonAction`, prioritized after the editor menu and before the expanded toolbar).
- Leaving the editor (`showHome`) closes the bar without refocusing the editor; every document open (including incoming share/open-with intents, which pass through the same wrapped `openEditor`) resets stale find-bar state. Muya itself already resets its match state on `setContent` (`searchResetOnSetContent.spec.ts`).
- Content edits re-run the open query on the existing `syncMarkdown` dirty path, so highlights and the counter never go stale.
- The floating selection toolbar is suspended while the bar is open; the query input guards IME composition on Enter.

## Deliberately out of scope (v1)

Replace, regex/case/whole-word toggles, search history, and cross-document search. Muya's engine already supports several of these if a later PR wants them.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] UI/UX change
- [ ] Android native integration
- [ ] CI/build/documentation

## Test plan

- [x] `pnpm test` — 356 passed (12 new `documentSearch` controller cases: open/close contract incl. `selectHighlight`, query feedback, wrap navigation, edit refresh, document-switch reset, missing-editor guards; plus `close-editor-search` Back-priority cases)
- [x] New e2e `mobile-editor-search.spec.ts` — 5 passed (real Muya engine on the Pixel 5 profile: highlight counts `1/3`, wrap-around `1→2→3→1` and `1→3`, no-match state with disabled navigation, close clears every highlight and returns focus to the editor, query switch/clear)
- [x] Affected e2e regression — 32 passed (editor toolbar, selection toolbar, markdown editing, home navigation, local draft lifecycle)
- [x] `pnpm typecheck`, `pnpm lint`, `pnpm build`
- [x] Emulator smoke (API 35, emulator 36.6.11) — real-device pass over CDP + adb input: typed content through the native key path, query `apple` → counter `1/3` with 1 active + 2 passive highlights, wrap navigation `1/3 → 2/3 → 3/3 → 1/3`; native Back #1 dismissed the soft keyboard (IME-consumed, standard Android), Back #2 closed the bar with zero stale highlights and the editor cursor restored onto the last active match (document.activeElement inside the editor). The earlier blocker was environmental: a winnat port reservation had swallowed the emulator console port range.

## Android notes

Web-layer only: no plugin, manifest, or Gradle changes. Back handling goes through the existing `getAppBackButtonAction` decision table with tests.

## Reviewer notes

`scrollIntoView({ block: 'center' })` targets the `.mu-highlight` span inside `muya.domNode` (the `editorHost` ref goes stale after Muya's init swap, so the DOM query uses the editor instance). The `1/1`-style counter reads Muya's returned `matches`/`index` directly rather than duplicating match state app-side.


Renakoni:
Both findings addressed in d2e53fa:

**1. Close focus fallback** — `closeSearch` now checks whether a match was active before the close. The active-match path keeps Muya's `selectHighlight` cursor restore unchanged; the empty-query and no-match paths fall back to `muya.focus()`, whose documented behavior restores the last caret while it still belongs to the current document (`editor/index.ts:357`), i.e. the editing position from before the search opened. The Android Back path reuses the same `closeSearch`, so it inherits the fallback. Coverage: two controller unit cases (empty query → `focus()` called; no-match query → `focus()` called), one guard case (`restoreCursor: false` leaves focus alone for editor exit), and two real-Muya e2e regressions asserting `document.activeElement` lands back inside the editor host for both flows.

**2. IME Enter** — replaced the `.prevent` modifier with an explicit `handleSearchEnterKeydown(event, findNext)` (exported from `documentSearch.ts` for testability): it returns before `preventDefault()` when `event.isComposing` is true, so a composing Enter is left entirely to the input method; a real Enter is default-prevented and navigates. Unit cases assert exactly the requested matrix (normal Enter: prevented + emitted; composing Enter: not prevented + not emitted), plus an e2e that presses a real Enter (1/3 → 2/3) and then dispatches a cancelable composing Enter, asserting `defaultPrevented === false` and no navigation. Note the `@keydown.enter` modifier also never fires for composition keydowns that report `key: "Process"`, which is the common Chromium path — the in-handler guard covers IMEs that report `key: "Enter"` with `isComposing: true`.

A real-device CJK composition smoke (Pinyin candidate confirmation) needs a Chinese IME installed on the AVD, which the stock image lacks; the synthetic-event e2e plus the unit matrix cover the contract. Full suite after the fix: 360 unit / 8 search e2e, lint + typecheck clean.

PR #99, opened 2026-07-11
Add mobile document outline with heading navigation

Renakoni:
## Summary

Adds a persistent **Outline** action to the editor top bar (Back | title | Search | Outline | More; hierarchical-list icon, distinct from the three-dot menu). Tapping it closes conflicting overlays, blurs editor focus so the soft keyboard retracts (Muya's cached internal selection is untouched), waits for the visual viewport to settle (event-driven `visualViewport` resize quiescence with a hard 450 ms bound — no fixed delay), snapshots a fresh `muya.getTOC()`, and opens a modal bottom sheet.

The sheet sizes itself to short outlines, caps at ~52dvh for long ones, keeps its header fixed while only the list scrolls (`overscroll-behavior: contain` — one active scroll region), respects the bottom safe-area, and leaves the editor visible behind the shared scrim but non-interactive. It closes on row selection, its explicit close button, scrim tap, and Android Back (new `close-editor-outline` decision, prioritized after the editor menu).

## Navigation mapping (the load-bearing part)

Muya's stable TOC `slug` is **not** stamped onto the heading DOM. `getTOC()` walks only top-level `scrollPage` children, which render as the direct children of `.mu-container` — so the flat TOC index maps 1:1 onto `querySelectorAll('.mu-container > h1..h6')`. Headings inside blockquotes, lists, or raw HTML are excluded by both sides of the mapping (verified: `state/getTOC.ts`, `scrollPage/index.ts:77`, `atxHeading/setextHeading` `tagName = h${level}`; same resolution the desktop `tocNavigation.ts` documents). Selection scrolls the real scroll container (`.editor-host-shell`) so the heading lands near the top with context; a stale slug or shrunken DOM resolves to null and the sheet just closes — no wrong-heading scroll, no Markdown change, no caret move, keyboard stays closed.

## Presentation

Rendered plain heading text from Muya (inline markdown/links/images already stripped). Indentation normalized from the shallowest level present and capped at 3 steps; top-level rows get weight only — no H1/H2 badges, no collapsible tree. Rows are 48 px minimum, clamp to two lines (`overflow-wrap: anywhere` for CJK/long tokens), `dir="auto"` per row for RTL titles. Empty documents keep the Outline entry and show a restrained empty state. Accessibility focus moves into the `role="dialog"` panel on open (DOM focus only — editing selection and keyboard untouched).

Fresh TOC per open; no synchronized global TOC store (the editor is non-interactive behind the sheet, so the snapshot cannot go stale except through document replacement, which resets the sheet). `muya-core.d.ts` already declared `getTOC()`; **no Muya internals changed**.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] UI/UX change
- [ ] Android native integration
- [ ] CI/build/documentation

## Test plan

- [x] `pnpm test` — 380 passed (24 new: projection normalization/cap/repeats, index mapping with blockquote+list decoys, stale-DOM null, viewport-settle quiet/extend/hard-bound/no-viewport, controller open sequencing, empty-TOC open, stale-target fail-safe, destroyed-editor abort, document-switch reset; plus `close-editor-outline` Back priority)
- [x] New e2e `mobile-editor-outline.spec.ts` — 6 passed against real Muya (top-level-only listing with quoted-heading exclusion, repeated-heading navigation to the correct occurrence with scroll-landing assertion, markdown untouched, scrim/close dismissal, empty state, focus dismissal + dialog focus)
- [x] `pnpm lint`, `pnpm typecheck`, `pnpm build`
- [x] Emulator smoke (API 35, emulator 36.6.11, CDP + adb input) — real keyboard opened (viewport 867→554), Outline tap retracted it and the sheet opened at the settled viewport (867), 5 rows with the quoted heading excluded, selecting the second "Beta" scrolled the correct occurrence into the top half of the editor with the keyboard still closed, native Back closed a re-opened sheet while staying in the editor
- Muya suite untouched (no vendored changes); no Android native changes

## Reviewer notes

`--outline-indent` is a runtime inline-style variable, added to the theme-token test's existing `RUNTIME_DEFINED_TOKENS` exemption alongside `--editor-area-width`. Deliberately out of scope per the plan: scroll-spy, collapse state, outline search, heading editing, sidebar/full-screen variants. Upward sheet expansion was skipped: the 52dvh cap plus internal list scrolling covers long outlines without a second gesture system.


Renakoni:
Both findings addressed in 253563c:

**1. Pending-open lifecycle** — `createDocumentOutline` now owns the async open explicitly: an `openPending` flag rejects duplicate concurrent opens, and a generation token invalidates the unfinished open whenever `closeOutline()` or `resetForNewDocument()` runs — both bump the generation even when the sheet is not yet open, which covers document replacement, editor exit (`showHome`), and competing surfaces. `openEditorSearch()` now calls `closeEditorOutline()` symmetrically, so your reproduced sequence (Outline tap → Search tap inside the settle window) cancels the pending outline instead of double-opening. After the viewport settles the request must still be current **and** `getEditor()` must return the same instance captured before the wait, closing the replaced-editor hole. Regression coverage: the three requested cases as controller unit tests (reset-during-settle never opens with `getTOC` unreached; duplicate open produces exactly one dismiss/settle/getTOC sequence; close-during-settle for the competing-surface path) plus an editor-instance-swap abort case, and two e2e cases driving the real timing (Outline+Search clicks in one task → after the full settle bound only the search bar exists; triple-tap → exactly one sheet).

**2. Modal focus isolation and restoration** — the background (both top-bar variants, the editor pane, the selection toolbar, and the editor toolbar) is now `inert` while the sheet is open, so hardware-keyboard and switch-access focus cannot reach behind the scrim; your Shift+Tab-to-toolbar repro is covered by an e2e that walks Shift+Tab plus eight forward Tabs and asserts focus never leaves the dialog (Tab additionally cycles inside the panel, since `inert` alone would still let focus fall out to BODY). Every close path — close button, scrim tap, Escape, Android Back (same `closeOutline`), and heading selection — restores DOM focus to the Outline top-bar trigger via a `watch` on `outlineOpen`, guarded by `isConnected` for editor-exit unmounts. The target is a plain button: focusing it cannot reopen the soft keyboard and does not touch Muya's editing selection (no `muya.focus()` involved). E2e asserts the button holds focus and `activeElement` is outside the editor host for all four close paths.

Full suite after the fixes: 384 unit (4 new races), 10 outline e2e (4 new), lint/typecheck clean. One pre-existing flake note: the rapid-tap e2e failed once on first run inside the full spec — the failure was in the shared `expectEditorReady` fixture (editor never left loading, 30 s), not in outline logic; it passed in isolation and the full spec passed 10/10 on re-run.

Renakoni:
Addressed in 7908d70. The mutual-exclusion rule is now applied at every competing-surface entry point in the App orchestration layer, not just Search:

- `toggleEditorMenuExclusive` and `toggleEditorToolbarExclusive` wrap the template bindings for the More button and the toolbar expand handle — both call `closeEditorOutline()` first, which cancels the pending generation as well as an open sheet. (Cancelling before a close-direction toggle is a harmless no-op, so the wrappers stay unconditional.)
- The link insert sheet workflow (`openLinkSheet` call site) closes the outline before opening — that covers the toolbar/selection-toolbar link command paths, the remaining editor-level modal reachable during the settle window.
- Back-path closers (`closeEditorMenu`/`closeEditorToolbar`) are close-direction only and need no wrapping.

Regression coverage for both requested cases: an e2e that clicks Outline + toolbar-expand in one task and asserts after the full settle bound that the states never coexist (`outline === 1 && toolbarExpanded` is false; the toolbar interaction wins and the sheet count is 0), and an e2e that clicks Outline + More in one task against a mocked shareable Android document (the More button only renders for shareable documents) asserting the action sheet stays visible and the outline sheet count is 0 after the settle bound.

Full suite after the fix: 384 unit, 12 outline e2e (2 new), 17 toolbar/menu regression e2e, lint/typecheck clean.

PR #101, opened 2026-07-12
Extend the floating selection toolbar with long-press caret states

Renakoni:
## What

A long-press inside the editor now opens the existing floating `MobileSelectionToolbar` — for a collapsed caret as well as a text selection, including in an empty document. The floating toolbar remains the only clipboard-action UI: the bottom formatting `MobileEditorToolbar` has a zero-line diff in this PR, and the native Android floating action toolbar stays suppressed.

| Editor state | Floating toolbar actions |
| --- | --- |
| Editable, no selection, clipboard has content | Paste, Select all |
| Editable, no selection, clipboard empty | Select all |
| Editable, selection, clipboard has content | Cut, Copy, Paste, Select all |
| Editable, selection, clipboard empty | Cut, Copy, Select all |
| Read-only, selection | Copy, Select all |
| Read-only, no selection | Select all |

Select all stays available in an empty document. Ordinary taps, ordinary caret placement, typing, and caret movement never open the toolbar; the existing selection-driven behavior (appearance, placement, dismissal, drag handles, live Muya selection, clipboard command pipeline) is preserved unchanged.

## How

**Long-press signal.** `MainActivity.onActionModeStarted` forwards the suppressed floating ActionMode start — the exact moment Android would have shown its own clipboard menu (long-press selection, double-tap word select, insertion-caret menu) — to the web layer as a `selectionContextRequest` event on the existing AndroidSelection plugin (~20 native lines). Web dev/e2e use an in-page `contextmenu` fallback, deliberately not installed on Android where the native signal covers every entry path.

**Caret sessions** (`selectionToolbarLongPress.ts`). A context request with a collapsed editor caret opens a caret session; with an existing selection it only refreshes the clipboard state (the selection-driven display already owns that case). Sessions follow the native insertion-menu conventions: they end on an ordinary tap (toolbar taps excluded), caret movement, edits, command completion, competing surfaces through the App orchestration layer, and document replacement; a caret widened into a selection via the insertion handle hands over to the unchanged selection-driven display. Entries are generation-guarded against surfaces opening during the async clipboard query.

**State table** (`selectionToolbar.ts`). `getSelectionToolbarCommands` now derives from `{hasSelection, canPaste, canWrite}`; Paste is truthful to actual clipboard content (native `readClipboardText` on Android; the web Clipboard API with a capability-optimistic fallback, refreshed on every context request and whenever a selection appears). `shouldShowSelectionToolbar` admits a collapsed caret only during a session, never overriding suspension/readiness/bounds; an all-zero collapsed-caret rect in an empty block falls back to the caret's block element for placement.

**Two caret-specific correctness details:** paste from a caret session adopts the live DOM caret into Muya's selection model (`syncMuyaSelectionFromDom`), so it lands at the long-pressed position instead of Muya's stale cached selection — the emulator smoke and a dedicated e2e prove the payload lands inside the long-pressed paragraph; and caret rows pass a null restore range so they can never resurrect an earlier expanded selection.

## Verification

- Unit: 425 passing (19 new — the state table exhaustively row by row, caret visibility gating, session enter/exit/handover rules, stale-entry cancellation).
- E2E: all specs passing (9 new — selection state-table with real clipboard content, caret paste landing inside the long-pressed paragraph, Muya's progressive select-all from a caret (block → document), empty-document select all, ordinary-tap/typing negatives, subsequent-tap dismissal, and the read-only rows driven through the Android mock's new native-signal channel — the same entry path real devices use). Existing selection specs pass without behavioral edits.
- Lint, typecheck, build.
- API 35 emulator smoke with REAL long-presses via adb (10/10): ordinary tap opens nothing; long-press selects a word with visible native drag handles and no native menu while the floating toolbar shows Cut/Copy/Paste/Select all; copy dismisses; a caret long-press at a paragraph end offers exactly Paste/Select all with a confirmed collapsed caret; paste inserts at the caret and dismisses; the bottom formatting toolbar is untouched throughout (screenshot-verified).

### DIFF
--- a/android/app/src/main/java/io/github/renakoni/marktextandroid/AndroidSelectionPlugin.java
+++ b/android/app/src/main/java/io/github/renakoni/marktextandroid/AndroidSelectionPlugin.java
@@ -69,6 +69,10 @@ public class AndroidSelectionPlugin extends Plugin {
         notifyListeners("selectionTap", data);
     }
 
+    void emitSelectionContextRequest() {
+        notifyListeners("selectionContextRequest", new JSObject());
+    }
+
     @PluginMethod
     public void performNativeSelectAll(PluginCall call) {
         String reason = call.getString("reason", "unspecified");
--- a/android/app/src/main/java/io/github/renakoni/marktextandroid/MainActivity.java
+++ b/android/app/src/main/java/io/github/renakoni/marktextandroid/MainActivity.java
@@ -98,6 +98,14 @@ public class MainActivity extends BridgeActivity {
         activeSelectionActionMode = mode;
         activeSelectionActionModeStartUptimeMs = SystemClock.uptimeMillis();
         recordSelectionActionModeEvent("action-mode-started", modeDescription(mode));
+
+        // The suppressed floating ActionMode starting is exactly the moment
+        // Android would have shown its clipboard menu (long-press selection,
+        // double-tap word select, insertion-caret menu). Hand that moment to
+        // the web layer so the app-owned toolbar can offer clipboard actions.
+        if (isSuppressedSelectionActionMode(mode)) {
+            notifySelectionContextRequest();
+        }
     }
 
     @Override
@@ -193,6 +201,20 @@ public class MainActivity extends BridgeActivity {
             builder.append(menu.getItem(index).getItemId());
         }
         return builder.toString();
+    }
+
+    void notifySelectionContextRequest() {
+        recordSelectionActionModeEvent("selection-context-request", "notify");
+
+        if (getBridge() == null) {
+            return;
+        }
+
+        PluginHandle handle = getBridge().getPlugin("AndroidSelection");
+        Plugin plugin = handle == null ? null : handle.getInstance();
+        if (plugin instanceof AndroidSelectionPlugin) {
+            ((AndroidSelectionPlugin) plugin).emitSelectionContextRequest();
+        }
     }
 
     void notifySelectionTap(float cssX, float cssY) {
--- a/src/App.vue
+++ b/src/App.vue
@@ -608,12 +608,6 @@ const {
   logger: editorLog,
 })
 
-const canWriteSelectionDocument = computed(() =>
-  documentState.value.autosaveTarget === 'android-document'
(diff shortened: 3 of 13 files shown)


Renakoni:
Both findings addressed in 092f2a7.

**[P2] Unwritable source URI ≠ read-only editor** — agreed, and the Save-copy lifecycle regression you cited makes the contract unambiguous. The `canWriteSelectionDocument` computed (mapping to `currentAndroidDocumentCanWrite`) is gone; `MobileSelectionToolbar` now derives editability from the editable surface itself — the Muya root's `contenteditable` state, read live in the same update pass that samples the selection. Every document currently opens editable in memory, so Cut and Paste stay available for unwritable-URI documents, and the read-only table rows activate automatically if a genuinely non-editable surface ever exists. The reworked e2e proves both sides on the native-signal path: a `canWrite: false` Android document offers `cut, copy, paste, selectAll` (selection) and `paste, selectAll` (caret); flipping the surface to `contenteditable=false` drops to `copy, selectAll` and `selectAll` per the table. The lifecycle spec (edit read-only doc → Save copy) passes unchanged.

**[P2] Stale caret behind a pending clipboard query** — the request is now bound to the caret that existed at the long-press: `enterFromContextRequest` captures the entry caret BEFORE awaiting the clipboard query and identity-checks it (node + offset) after resolution, so a caret that moved through an ordinary tap, typing, or arrow keys never receives a toolbar. Complementing that, `endCaretSession` advances the generation (and `notifyDocumentEdited` is now unconditional rather than gated on an active session), so every dismissal path — competing surfaces, edits, command runs, document replacement — invalidates a request whose toolbar never became visible. Four new deferred-query regressions cover exactly the orderings you described: caret moved mid-query, edit mid-query, surface opened mid-query (all remain hidden), plus the unchanged-caret happy path still opening.

Verification: 429 unit, 110 e2e (incl. `mobile-android-document-lifecycle.spec.ts` untouched and green), lint/typecheck, and the API 35 real-long-press smoke rerun on the rebuilt APK (10/10 — ordinary tap opens nothing, caret long-press still yields exactly Paste/Select all, bottom toolbar untouched).

PR #117, opened 2026-07-13
Ask before an incoming Intent discards an unpreservable document

Renakoni:
## Problem

An incoming Android Intent replaced the live document even when preserving it failed. **[P1 — data safety / document lifecycle]**

`preserveCurrentDocumentBeforeIncomingOpen()` returned no result: it awaited `saveAndroidDocument()`, optionally attempted a recovery draft, and returned. Both incoming handlers then **unconditionally** continued to `openAndroidDocumentResult()` or shared-text replacement. Local-draft preservation had the same gap because `saveDraft()` converted storage failures into UI state but returned nothing.

So the current editor could be silently replaced — losing unsaved work — after:

- the source file was moved or deleted (Android save fails with `FileNotFoundException`);
- an Android document save otherwise failed;
- "Save failed drafts" (recovery drafts) was disabled;
- local drafts were disabled;
- `localStorage` persistence failed;
- or a second save arrived while another Android save was in flight.

An unsolicited external Intent is not a deliberate "leave this document," so dropping unsaved work here violated the product's **"never lose a word"** bar.

## Fix — ask before discarding (honors settings; no silent override)

Make preservation report its outcome and only replace the editor after it durably completes.

- `preserveCurrentDocumentBeforeIncomingOpen()` now returns `preserved | blocked`.
  - `saveDraft()` returns whether the local draft is durably safe (persisted, or there was nothing to keep).
  - `persistAndroidRecoveryDraft()` returns whether it actually wrote, and no longer throws on a storage failure.
  - An Android document is `preserved` when it saved, had nothing to lose, or a recovery draft was durably kept — otherwise `blocked`.
- When preservation is **blocked**, the editor stays mounted and a calm `IncomingOpenPrompt` asks: **Keep editing** (default, safe) or **Discard & open**. This respects the user's recovery/local-draft settings rather than silently overriding them, and mirrors the app's existing unsaved-changes exit prompts. **Android Back** maps to "keep editing" above every other layer.

The **happy path is unchanged**: when preservation succeeds the incoming document opens with no prompt.

### Why ask instead of always keeping a silent recovery draft

The crash-recovery-draft direction is unconditional precisely because a crash can't prompt. Here the app is foregrounded and the user just interacted, so we *can* ask — and asking honors a user who deliberately turned draft persistence off, instead of writing a draft they opted out of.

## Design note

`IncomingOpenPrompt.vue` reuses the existing exit-prompt styles/pattern. The "Keep editing" action is visually primary (the safe default); "Discard & open" is the danger action. If the user keeps editing, the incoming document is dropped and can be re-triggered from the source app. ("Save a copy…" is a natural fast-follow, deliberately left out of this PR.)

## Verification

- Unit (`vitest run src`): **491 passed** — added orchestration `blocked`/`discard` paths, persistence durability signals (`saveDraft` / `persistAndroidRecoveryDraft`), and the Back-priority of the new prompt.
- `npm run typecheck` (vue-tsc -b): clean.
- `npm run lint` (eslint, `--max-warnings 0`): clean.
- Focused e2e `mobile-share-intent.spec.ts`: **9/9 passed** (happy-path preservation unbroken).

## Files

10 changed (+378 / −33) + new `IncomingOpenPrompt.vue`: orchestration result + handlers, persistence durability returns, App state/handlers/Back wiring, `EditorScreen` props/events, `appExitDecisions` Back action, en/zh i18n.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1541,6 +1541,19 @@ function confirmIncomingOpenAfterBlockedPreserve(request: {
   incomingName: string
   proceed: () => Promise<void>
 }) {
+  // The safety prompt must be the only surface on screen so it is visible and
+  // Android Back answers it — otherwise an open menu/outline (z-index 40) would
+  // cover it. Close transient editor chrome WITHOUT touching the editor, which
+  // stays mounted with its unsaved content.
+  closeEditorMenu()
+  closeEditorToolbar()
+  closeEditorOutline()
+  closeEditorSearch({ restoreCursor: false })
+  closeLinkSheet()
+  closeTableSheet()
+  draftExitPromptOpen.value = false
+  androidExitPromptOpen.value = false
+
   pendingIncomingOpenProceed = request.proceed
   incomingOpenName.value = request.incomingName
   incomingOpenPromptOpen.value = true
@@ -1557,6 +1570,8 @@ function keepEditingInsteadOfIncomingOpen() {
     documentState.value.autosaveTarget === 'android-document'
       ? getAndroidEditorStatus()
       : 'Edited'
+  // Resume the queue: any intent that arrived while the prompt was open runs now.
+  void incomingDocuments.resolveIncomingDecision()
 }
 
 async function discardCurrentAndOpenIncoming() {
@@ -1567,6 +1582,8 @@ async function discardCurrentAndOpenIncoming() {
   if (proceed) {
     await proceed()
   }
+  // Drain after the replacement so a queued intent preserves the new editor.
+  void incomingDocuments.resolveIncomingDecision()
 }
 
 function requestLifecycleFlush(reason: string) {
--- a/src/App.vue
+++ b/src/App.vue
@@ -1579,11 +1579,21 @@ async function discardCurrentAndOpenIncoming() {
   pendingIncomingOpenProceed = null
   incomingOpenPromptOpen.value = false
   incomingOpenName.value = ''
-  if (proceed) {
-    await proceed()
+  try {
+    if (proceed) {
+      await proceed()
+    }
+  } catch (error) {
+    const message = getAndroidDocumentUserMessage(error)
+    homeNotice.value = message
+    status.value = message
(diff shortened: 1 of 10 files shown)


Renakoni:
All four findings addressed in `2bfb3ef`.

**[P1] Recover the latest editor snapshot after the awaited save.**
The payload is now read AFTER `await saveAndroidDocument()`: the code re-flushes (`syncDocumentFromEditor(..., true)`) and re-reads the current content, so a `changed-during-save` result keeps the newest text rather than the pre-save snapshot. Regression test: with a save that mutates the document to `# Edited during save` and returns `false`, the recovery payload is now `# Edited during save`.

**[P1] Serialize incoming-open transactions while a decision is pending.**
Incoming opens run one at a time through a queue (`scheduleIncomingTransaction` → `drainIncomingTransactions`). While a blocked prompt awaits the user, `awaitingDecision` freezes the queue, so a newer intent neither opens behind the prompt nor runs its own preservation concurrently (latest-wins). `resolveIncomingDecision()` — called by App on both Keep and Discard — resumes the queue, so a queued intent then preserves the *current* editor. Regression test: a second intent delivered during a pending prompt makes no `openAndroidDocumentResult` call and does not run a second preservation until the first decision resolves.

**[P2] Show the preservation prompt above existing editor chrome.**
`confirmIncomingOpenAfterBlockedPreserve` now closes transient editor chrome (menu, toolbar, outline, search, link/table sheets) — without touching the editor — before showing the prompt, so it is the only surface and Android Back answers it rather than a hidden sheet.

**[P2] Make the new modal own keyboard and accessibility focus.**
`IncomingOpenPrompt` moves focus to **Keep editing** on open, traps Tab within its two actions, maps Escape to the safe keep action, and restores focus on close. `EditorScreen` marks the editor, header, toolbars, and overlays `inert` while the prompt is open — following the existing focused-sheet pattern, no new dialog framework.

Verification: full unit suite 493 passed (incl. the two new regressions), typecheck + lint clean, and the focused `mobile-share-intent` e2e 9/9 (happy path unaffected by the serialization).

PR #160, opened 2026-07-16
Render authored soft breaks as <br> on export (fixes #143)

### LINKED ISSUE #143: Soft line breaks are lost in HTML/PDF export
The editor shows a soft line break (a bare \`\n\` inside a paragraph) as a line break (\`.mu-content\` is \`white-space: pre-wrap\`), but HTML/PDF export collapses it to a space — the two lines run together.

This is upstream marktext/marktext#3676, fixed upstream by marktext/marktext#4844 with a presentation-layer CSS rule: the exported HTML keeps the CommonMark-conformant \`\n\` and the export stylesheet renders it with \`white-space: pre-wrap\` on paragraphs and tight list items. An earlier local port of that fix was never merged and is stale; it should be redone against current main.

**Scope:** one CSS rule in the vendored export stylesheet plus a conformance test (soft break stays \`\n\`, never \`<br>\`; a real hard break still renders \`<br>\`).

Split from #142 ("Related defect" section). Independent of the deferred soft-break-as-space preference tracked there.

Renakoni:
Fixes #143 — soft line breaks lost in HTML/PDF export (upstream marktext#3676).

## The final design (review rounds 1-6)

The first approach ported upstream marktext#4844 verbatim: `white-space: pre-wrap` on export containers. Six review rounds proved that presentation-layer approach unrepairable — a container holds authored soft breaks (should break), authored raw-HTML formatting (should collapse), and marked's serializer pretty-printing (should be invisible), and neither CSS nor post-hoc DOM inspection can tell them apart. Every refinement produced a new counterexample: phantom rows, deleted `&nbsp;`, collapsed raw `<p>`/`<section>`/`<menu>`, broken inline-restyling themes.

The fix now lives at the only layer that knows: **the inline text leaf of the parser**. `getHighlightHtml` gains `exportSoftBreaks` (used only by `MarkdownToHtml`, i.e. the HTML/PDF export): the leaf text renderer replaces its newlines with `<br>` (global-regex `replace`; muya's build targets chrome70, which lacks `replaceAll`) — CommonMark explicitly permits rendering soft breaks as hard line breaks. Code spans, raw HTML, and hard breaks are separate token types. Protection is ONE authoritative state with two tiers: a raw-text slot for `script`/`style`/`title`/`textarea` (same-name close only) and a nesting stack for the ordinary containers `pre`/`code`/`kbd`. Transitions are scanned from every html token, BLOCK tokens included (type-6 blocks end at a blank line and can carry unbalanced tags); self-closing-style opens and closes count like plain forms; marked's mis-trackable `escaped` boolean is not trusted — each leaf's `escaped` is recomputed from this state. Everything else is byte-for-byte marked-canonical; all export-stylesheet whitespace rules are removed.

## Verification

- Unit (`softBreakExportHtml.spec.ts`, 28 tests, incl. raw-text content for all seven tags, bidirectional state-leak cases, mismatched closes, self-closing opens and closes, same-name `<code>` nesting, and unbalanced block-token tags): soft → `<br>` in paragraphs, tight items, between inline-wrapped lines, inside inline spans; hard breaks unchanged; option off keeps the spec newline; raw HTML, raw blockquotes, inline comments, `&nbsp;`, code spans, `<p>[TOC]</p>`, and theme separators all byte-canonical.
- Browser layout (`tests/e2e/export-soft-break-layout.spec.ts`, 4 tests, real Chromium against the real stylesheet; the raw-HTML case measures CONTAINER heights, whose sensitivity to a simulated pre-wrap regression was verified): soft breaks measure as true second lines; canonical serializer newlines create no phantom rows; raw HTML (`<p>`, `<div>`, `<section>`, `<menu>`) keeps normal single-line layout; an inline-restyling theme keeps its separating space.
- Muya app-gate **1505/1505** (post-rebase onto main with #159/#161 merged); root **581/581** (incl. the vendor-sync contract); `pnpm typecheck`, `pnpm lint`, `pnpm build` clean; `node_modules/@muyajs/core` resynced.

Kept in lockstep with upstream marktext#4951 (same commit content; plus a real-Chromium e2e there for the sanitize path happy-dom cannot reproduce).

### DIFF
--- a/src/features/editor/documentOutline.test.ts
+++ b/src/features/editor/documentOutline.test.ts
@@ -18,9 +18,9 @@ describe('projectOutlineItems', () => {
 
   it('normalizes indentation from the lowest heading level present', () => {
     const items = projectOutlineItems([
-      { content: 'Alpha', lvl: 3, slug: 'a' },
-      { content: 'Beta', lvl: 4, slug: 'b' },
-      { content: 'Gamma', lvl: 3, slug: 'c' },
+      { content: 'Alpha', lvl: 3, slug: 'a', index: 0 },
+      { content: 'Beta', lvl: 4, slug: 'b', index: 1 },
+      { content: 'Gamma', lvl: 3, slug: 'c', index: 2 },
     ])
 
     expect(items.map(item => item.indent)).toEqual([0, 1, 0])
@@ -29,9 +29,9 @@ describe('projectOutlineItems', () => {
 
   it('caps deep hierarchies at MAX_OUTLINE_INDENT steps', () => {
     const items = projectOutlineItems([
-      { content: 'H1', lvl: 1, slug: 'a' },
-      { content: 'H5', lvl: 5, slug: 'b' },
-      { content: 'H6', lvl: 6, slug: 'c' },
+      { content: 'H1', lvl: 1, slug: 'a', index: 0 },
+      { content: 'H5', lvl: 5, slug: 'b', index: 1 },
+      { content: 'H6', lvl: 6, slug: 'c', index: 2 },
     ])
 
     expect(items[1].indent).toBe(MAX_OUTLINE_INDENT)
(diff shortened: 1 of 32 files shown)


Renakoni:
Closing pending the export-pipeline refactor (upstream marktext#4954; local blueprint in refs/export-pipeline-refactor-design-2026-07-16.md). The inline-token `<br>` approach is likely the right core, but the raw-HTML guard kept requiring renderer-level HTML-semantics emulation — that belongs in the restructured pipeline. Branch kept for reference; it carries the 28-test regression suite from the review rounds. #143 stays open.

PR #163, opened 2026-08-04
Import non-persistable incoming documents as local drafts

### LINKED ISSUE #162: Open-with documents leave no trace: files opened from other apps disappear after exit
## Report

Opening a Markdown file from another app (for example a file received in QQ) through the system "Open with" chooser works, but the document leaves no trace in MarkText: after leaving the app there is no entry on the Home screen, and the file is effectively gone. Users expect an opened document to appear in their history.

## Root cause

Android delivers open-with / share-stream files as one-shot `content://` grants. App-provided URIs (e.g. QQ's FileProvider) never carry `FLAG_GRANT_PERSISTABLE_URI_PERMISSION`, so the grant dies with the task and can never be reopened. The app currently reacts by deliberately dropping the document from history:

- `AndroidDocumentsPlugin.emitOpenWithDocument` reports `persisted: false` for these grants.
- `createAndroidOpenWithDocumentEventAction` / `createAndroidShareDocumentEventAction` map that to `remember: false`.
- `createAndroidDocumentOpenResult` then returns `rememberDocument: null`, so no Recents record is written ("Opened temporarily").
- `documentStorage` additionally filters the recents store to `kind === 'android-document'`, and `openDocument` treats any other kind as "not openable yet" — the reserved `incoming-document` kind was never wired.

"Don't remember what we can't reopen" avoids dead history entries, but it silently discards the user's document instead of retaining the one durable thing available: the content.

## Fix

Import non-persistable incoming documents (open-with and shared files) as named local drafts at open time, the same way shared text is already imported:

- The file's content and display name become a local draft immediately, so it appears on Home, survives app exit, and reopens offline.
- Reopening the same unchanged file reuses the existing draft instead of multiplying copies.
- Edits autosave into the draft (durably) instead of into a URI that dies with the task.
- Persistable documents (the in-app picker, providers that allow persistable grants) keep the current real-document Recents behavior.
- When the "keep local drafts" setting is off, the current temporary-access messaging is preserved unchanged.

Behavior note: a transiently writable open-with session previously wrote edits back to the source file until the task ended; with the import model such edits stay in the app copy. Editing the original file remains the in-app picker's job, which holds a persistable grant.

Renakoni:
Fixes #162.

## Problem

A Markdown file opened from another app (for example a file received in QQ) opens fine, but leaves no trace in MarkText: after leaving the app it is not on the Home screen and cannot be reopened. Same for files shared into the app as streams.

## Root cause

These intents deliver one-shot `content://` grants without `FLAG_GRANT_PERSISTABLE_URI_PERMISSION`, so the URI dies with the task. The app reacted by deliberately not recording the document (`persisted: false` → `remember: false` → `rememberDocument: null`, status "Opened temporarily"), and the storage layer additionally filters recents to `kind === 'android-document'`. "Don't remember what we can't reopen" avoided dead history entries — by silently discarding the document instead.

## Fix

The one durable record such a grant can never be is the content itself, so a non-persistable incoming file is now **imported as a named local draft at open time**, the same path shared text already uses:

- The file appears on Home immediately, survives app exit, and reopens offline through the existing draft machinery (rename, delete, share, save-to-device all work).
- Edits autosave durably into the draft instead of into a URI that dies with the task; the Android recovery-draft workaround is no longer involved in this flow.
- Reopening the same unchanged file reuses the existing draft (matched by display name plus normalized content) and bumps it to the top; a changed file becomes a new draft so neither version is lost.
- Persistable documents (in-app picker, providers granting persistable access) keep the current real-document Recents behavior.
- With the "keep local drafts" setting off, the previous temporary-access messaging is preserved unchanged.

`createAndroidDocumentOpenResult` loses its temporary-access limbo — it now only ever serves durable documents — and the orchestration routes on the `persisted` flag instead of threading `remember` through the open path. New status message "Imported this file as a local draft." is localized in all 10 locales.

## Behavior change to note

A transiently writable open-with session previously wrote edits back to the source file until the task ended (and then silently lost access). With the import model those edits stay in the app copy. Editing the original file remains the in-app picker's job, which holds a persistable grant. Flagging this explicitly for review.

## Verification

- `pnpm exec vue-tsc --noEmit`: clean.
- `pnpm test`: 75 files / 587 tests passed (includes 5 new workflow tests and 4 new orchestration tests covering import, draft reuse, changed-file fork, persisted routing, and the drafts-disabled fallback).
- `pnpm exec playwright test mobile-android-document-lifecycle mobile-share-intent`: 17 passed, including the rewritten warm open-with, temporary-exit, and shared-stream specs asserting the imported draft in storage.
- `eslint` on all changed files with `--max-warnings 0`: clean (repo-wide `pnpm lint` currently trips over the gitignored local-archive/ evidence folder, which CI's clean checkout does not have).
- `pnpm build`: passes.

Not run: on-device intent injection (emulator smoke); the mocked bridge e2e covers the event contract end to end.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -203,6 +203,7 @@ const SHARE_TEMPORARY_ACCESS_MESSAGE =
 const RENAME_TEMPORARY_ACCESS_MESSAGE =
   'Renamed, but Android only kept temporary access. Reopen it from Android to add it back to Recents.'
 const SHARED_TEXT_IMPORTED_MESSAGE = 'Imported shared text as a local draft.'
+const INCOMING_FILE_IMPORTED_MESSAGE = 'Imported this file as a local draft.'
 const ANDROID_SAVE_RECOVERY_MESSAGE = 'Save failed. A local recovery draft was kept.'
 const ANDROID_EXIT_RECOVERY_MESSAGE = 'Unsaved changes were kept as a recovery draft.'
 const ANDROID_EXIT_DISCARD_MESSAGE = 'Unsaved changes were discarded.'
@@ -1270,22 +1271,15 @@ function rememberAndroidDocument(document: OpenedAndroidDocument) {
 
 async function openAndroidDocumentResult(
   document: OpenedAndroidDocument,
-  options: { source?: AndroidDocumentOpenSource; remember?: boolean } = {},
+  options: { source?: AndroidDocumentOpenSource } = {},
 ) {
   const openResult = createAndroidDocumentOpenResult(document, {
     source: options.source,
-    remember: options.remember,
-    openWithTemporaryAccessMessage: OPEN_WITH_TEMPORARY_ACCESS_MESSAGE,
-    shareTemporaryAccessMessage: SHARE_TEMPORARY_ACCESS_MESSAGE,
     logger: androidDocumentLog,
   })
 
   homeNotice.value = openResult.homeNotice
-  if (openResult.rememberDocument) {
-    rememberAndroidDocument(openResult.rememberDocument)
-  } else {
-    protectAndroidDocumentImportedImages(document)
-  }
+  rememberAndroidDocument(openResult.rememberDocument)
   promptLocalDraftSaveOnExit.value = openResult.promptLocalDraftSaveOnExit
   currentAndroidDocumentCanWrite.value = openResult.currentAndroidDocumentCanWrite
   documentState.value = openResult.documentState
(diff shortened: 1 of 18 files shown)


Renakoni:
Review findings addressed in 425f309:

- **P1 (stored-draft deletion via reused id):** confirmed — stored drafts stay loaded in memory while the local-drafts setting is off (`readStoredLocalDrafts()` runs unconditionally at mount), and `discardLocalDraftAndShowHome` writes storage unconditionally. Deduplication now runs only when the draft will actually be persisted; with persistence off the session always gets an independent fresh id, so Discard can never remove a stored draft. Covered at both the workflow level (`never reuses a stored draft while local drafts are disabled`) and the orchestration level (`never adopts a stored draft id while local drafts are disabled`).
- **P2 (imported-image protection):** confirmed — the replaced temporary path called `protectAndroidDocumentImportedImages`, the import branch did not. The orchestration now takes a `protectIncomingDocumentImages` callback (wired to the same registry protection) and invokes it for both open-with and shared-stream imports, before any draft handling and regardless of the drafts setting. Cleanup scans current document + drafts + recents, so the protection registry is what keeps these images alive once the imported draft is deleted.
- **P2 (empty files):** confirmed — `normalizeLocalDrafts` drops empty content. Empty or whitespace-only files now open as a plain named session: status `Ready`, nothing persisted, no import claim; the session becomes a durable named draft the moment content is typed (draft autosave preserves the custom display name). Regression tests at both levels.

Verification after the fixes: vue-tsc clean; `pnpm test` 75 files / 592 tests passed (5 new); the two affected Playwright specs 17/17; eslint clean on changed files.

PR #166, opened 2026-08-04
Port the desktop light themes with a catalog-driven theme system

Renakoni:
Part 1 of the desktop theme port (tracking plan: port every remaining desktop theme with real mobile adaptation; part 2 will bring the dark set).

## What

**Eight light themes ported from the desktop reference**, each read in full from its desktop `.theme.css` (+ its `prismjs/*.theme.css` code palette) and re-derived for the mobile token contract:

- **Classic Light** — the classic desktop default (green `#21b56f`, white paper, neutral headings; lives in desktop's base `:root`, not a theme file)
- **Catppuccin Latte**, **Everforest Light**, **Gruvbox Light**, **Rosé Pine Dawn**, **Solarized Light**, **Tokyo Night Light**, **Ulysses**

Ulysses keeps its desktop structural signatures as mobile-safe scoped rules: centered theme-blue headings, the filled blockquote panel, and the centered half-width dashed divider. Its rotated-checkbox animation is deliberately not ported (it would fight the mobile touch-target styling).

## Mobile adaptation rules (applied consistently)

- **WCAG AA**: every `--accent` is contrast-checked against its paper and deepened where the heritage hue fails (everforest green sat at 2.69:1 → olive; latte blue 4.34:1 → nudged; tokyo navy already passes and keeps the exact hue). `--theme-color` always keeps the heritage hue for editorial accents. Body ink fixed where desktop fails AA (solarized base00 → base01; ulysses' 0.7-alpha fade → full-opacity heritage gray).
- **Elevation reads lighter on a phone**: floats/sheets raise toward a lighter surface instead of desktop's darker panel tints.
- **Dark scrims** (desktop used light masks), `strong`/`em` keep the ink color, plain hex/rgba only (old-WebView safe), per the existing product rules.
- Per-theme Prism palettes ported faithfully (7 files; Classic Light uses Muya's bundled default, which *is* its palette).

## Infrastructure

- **`THEME_CATALOG`** is now the single source of truth: appearance settings derive the valid ids, the theme runtime derives the palette registry + dark set (the mapping table is gone — catalog ids are palette ids), and the picker derives its options. Adding a theme = one palette file + one catalog entry.
- **Picker upgrade**: theme names are product names shown verbatim in every locale (the two per-theme locale keys are replaced by translated Light/Dark group headings ×10 locales); the sheet renders palette swatch dots (paper · accent · heading · ink) per theme and on the closed trigger.
- **`themeTokens.test.ts`** now discovers palettes from the directory and adds two invariants: exactly one palette per registered theme, and each palette's `color-scheme` matching its registered appearance. Token-set parity and the completed-task AA check now run across all 12 palettes.

## Verification

- `pnpm typecheck` clean; `pnpm test` 75 files / 594 tests passed; eslint clean on changed files; `pnpm build` passes.
- New `mobile-theme-gallery.spec.ts` (11 tests, all passing): per theme, opens a rich fixture (headings, emphasis, quote, tasks, code, table, divider, link), asserts `data-theme` + resolved paper color, and saves a screenshot; plus a picker walk asserting all labels, swatches, and both group headings.
- Screenshots reviewed by hand (Ulysses signatures, gruvbox/tokyo papers, picker sheet).
- Full-device emulator screenshot acceptance for all themes is planned as one pass after the dark batch lands, to avoid three build/install cycles.


Renakoni:
Review findings verified and fixed in 0469dc8 (both were real):

- **Group semantics + heading contrast:** options now render inside real listbox groups (`role=group` with the heading as `aria-label`; the visible heading stays presentational), and the heading uses the primary ink — measured across all palettes, the faint token sat between 1.77:1 and 4.03:1 on the raised sheet, and even the muted token fails AA in 7 palettes, so `--text` (which a new token test now guarantees ≥4.5:1 on raised surfaces, added on the part-2 branch) is the only safe choice. Size, tracking, and case keep it subordinate.
- **Scroll restore:** opening the sheet now scrolls the selected option into view (centered), so a theme picked near the bottom of the 31-entry list no longer requires re-scrolling on every open.

PR #168, opened 2026-08-04
Port the desktop dark themes, part 2 of 2

Renakoni:
Part 3 (final) of the desktop theme port — stacked on #167; merge #166 → #167 → this.

## What

**The last ten dark themes ported from the desktop reference**, each read in full from its desktop `.theme.css` and `prismjs/*.theme.css` and re-derived for the mobile token contract:

Nightfox, Nord, Oxocarbon Dark, Palenight, Rosé Pine, Rosé Pine Moon, Solarized Dark, Synthwave '84, Tokyo Night, Tokyo Night Storm.

**This completes the port: every theme in the desktop registry now ships on Android** — 2 fixed modes + 31 custom themes (9 light, 22 dark), all catalog-driven.

## Mobile adaptation (same rules as #166/#167)

- One-dark elevation pattern (lighter raised surfaces, accent-strong lightens, dark scrims); 0.6-alpha desktop search highlights softened to ~0.3.
- Contrast-checked per theme: Solarized Dark's blue (4.08:1 on base03) and red (3.25:1) lift past WCAG AA for the accent/danger roles; the nightfox rose (3.64:1), nord aurora red (3.05:1), and palenight red (4.37:1) lighten for the danger role; heritage values always stay on the Muya bridge and headings. Everything else clears AA as-is (tokyo blue 5.8–6.8:1, iris 7.5–8.4:1, neon pink 6.7:1…).
- Tokyo Night and Storm share one desktop prism token set (scoped per theme); Storm keeps its desktop trait of a code panel darker than the paper.

## Verification

- `pnpm typecheck` clean; `pnpm test` 75 files / 594 tests passed — the directory-driven theme tests now cover all 33 palettes (token parity, completed-task AA, data-theme scoping, color-scheme agreement).
- `mobile-theme-gallery.spec.ts`: 32 tests passed (31 themes + the picker walk with both group headings); screenshots saved per theme, Synthwave '84 and Solarized Dark hand-reviewed.
- Next and final step (after the stack merges): one emulator pass installing the debug build on MarkText_API_35 and screenshotting themes on-device, with the MuMu adb-server protection protocol (reuse the running adb binary, never kill-server).

PR #169, opened 2026-08-05
Shard and parallelize the Mobile WebView E2E job

Renakoni:
## Why it was slow

Three compounding causes, confirmed from the config and run history (one recent run hit the 15-minute job timeout exactly):

1. **`workers: 1` pinned globally** in playwright.config.ts — 24 spec files ran strictly one at a time while the runner has four cores.
2. The suite has grown (172 tests; the theme gallery alone adds 30+ full page loads), and CI `retries: 2` triples the cost of any flake.
3. Every run re-downloads ~120 MB of Playwright Chromium.

## What changed

- **Workers**: `'100%'` in CI (one per core); local runs stay serial — a long-lived workstation dev server ages under parallel cold-transform load, which is exactly the flake the config's retry comment describes.
- **Sharding**: the job becomes a 2-shard matrix (`--shard=1/2`, `2/2`), so two runners each take half the suite; `fail-fast: false` keeps one shard's failure from masking the other's result.
- **Browser cache**: `~/.cache/ms-playwright` keyed on the lockfile; `playwright install` becomes a no-op on hit (`--with-deps` still applies system packages).

Expected wall time: from >15 min (timeout) to roughly **3–5 min per shard**, running concurrently.

## Verification

- `--shard=1/2` / `2/2` `--list`: clean 12+12 file split (99 + 73 tests).
- Full local run at 4 workers: **172 passed, 0 flaky, 4.8 min** — validates that tests are parallel-safe (each test uses a fresh context and storage against the one shared dev server).

Note: the open theme-stack PRs (#166–#168) run the workflow from their own branches, so they keep the old serial job until this merges and the stack is refreshed from main (happy to do that refresh once this lands).

### DIFF
--- a/.github/workflows/web-quality.yml
+++ b/.github/workflows/web-quality.yml
@@ -68,17 +68,31 @@ jobs:
         run: pnpm build
 
   e2e:
-    name: Mobile WebView E2E
+    name: Mobile WebView E2E (${{ matrix.shard }}/2)
     runs-on: ubuntu-latest
     timeout-minutes: 15
+
+    # The suite outgrew a single serial job (it started hitting the
+    # 15-minute timeout), so it runs sharded across two jobs with parallel
+    # workers inside each (see workers in playwright.config.ts).
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2]
 
     steps:
       - uses: actions/checkout@v4
 
       - uses: ./.github/actions/setup
 
+      - name: Cache Playwright browsers
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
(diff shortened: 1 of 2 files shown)


Renakoni:
Finding verified and fixed in facdebe — it was real: pnpm ≥7 forwards a bare `--` to the underlying command, so Playwright received it as a positional and the shard flag never took effect (local repro was even uglier: `--list` also gets swallowed and the invocation runs the whole suite). The workflow now invokes the binary directly.

CI logs from run 30971004420 now show the intended behavior:

- **Mobile WebView E2E (1/2): `Running 99 tests using 4 workers` — success, 5 min**
- **Mobile WebView E2E (2/2): `Running 73 tests using 4 workers` — success, 2 min**

Wall time for the e2e gate: **~5 minutes** (shards run concurrently), down from >15 (timeout).

PR #171, opened 2026-08-05
Add find and replace to the in-document search bar

### LINKED ISSUE #145: Find and replace
The in-document search bar only finds; a writing tool needs replace.

**Evidence:** Muya already exposes a tested public \`replace(value, { isSingle, isRegexp })\` covering single/all/regex replacement — the app simply never surfaces it.

**Minimal scope:** add a replace input plus "replace" and "replace all" to the existing search bar; reuse the existing match count, debounce, scrolling, and close-time cursor restore. No new search engine. Regex UI only if the product ever needs it.

**Done when:** replace is a single undoable edit; count/active match stay correct after replacing; empty value, no match, IME input, cross-block matches, replace-all, and 100+ matches are covered by unit/E2E tests.

Renakoni:
Closes #145.

## Design

The find bar is untouched for pure-find users. Its trailing edge gains one element — a find-replace toggle — that expands a second row **in place** under the find field (the top bar grows one row; never a sheet, never a settings detour):

```
┌──────────────────────────────────────┐
│ ←   Find in document   3/12   ↑  ↓  ⇄│
│ Aa   Replace with…      Replace  All │   <- only while expanded
└──────────────────────────────────────┘
```

- **Aa** toggles case sensitivity and re-runs the query immediately. It sits where Word mobile puts its settings gear, but it *is* the option — no sheet in between. Default is insensitive (mobile keyboards auto-capitalize).
- **Replace** consumes the active match and advances; Enter in the replace field repeats it, with the same IME-composition guard as the find field. **All** replaces everything and reports "Replaced N" in the count slot until the next search action.
- Quiet by design: text actions in `--accent-strong` (covered by the theme contrast contract), disabled at zero matches, expand toggle tints while open, focus moves into the replace field on expand and back to find on collapse.
- **Whole-word is deliberately absent**: regex `\b` word boundaries are Latin-centric — CJK characters are not `\w`, so "whole word" never does what Chinese/Japanese text expects. The engine supports it; the UI stays honest.

## Engine (vendored Muya)

- `Search.replace()` single-replace now skips matches the replacement itself contains. Previously `cat -> wildcat` left the active match *inside* the fresh insertion, so repeated Replace taps compounded the same spot forever. The classic advance/step-back behavior for inert replacements is regression-pinned.
- `replace()` accepts the full `ISearchOption` set (case sensitivity previously could not reach the post-replace re-search).

## Undo contract

Every user-visible replace is exactly **one undo step**. The history coalesces ops in a 1s window, so the controller wraps each replace in `history.cutoff()` + `jsonState.flush()` + `cutoff()`. New Muya specs pin both directions: a replace-all reverts wholesale in one undo, and two quick single replaces stay two separate undo steps. The controller also mutes its own synchronous `json-change -> refreshAfterEdit` echo, which would otherwise reset the just-advanced active match.

## Verification

- Muya search specs 15/15 (4 new: insertion-skip x2, undo boundary x2)
- `documentSearch` unit tests 26/26 (9 new); full unit suite 605/605; typecheck clean; changed files lint clean
- E2E `mobile-editor-search` 12/12 (4 new: expand/collapse focus handoff, replace-one + Enter advance, replace-all + notice + disabled states, case toggle)
- 10 locales gain the six new strings (key parity kept)


Renakoni:
**Verification of review round two** — both findings are real; fixed in ce8bb72.

**Tail compounding (P2 — confirmed, arithmetic checks out exactly).** With `cat dog cat` and `cat -> wildcat`: after the second replace, `next = min(index + inside, total - 1) = min(2, 1) = 1` — the entry at 1 is the match inside the *just-inserted* second wildcat, and a third tap produced `wildwildcat`. The regression test indeed stopped one step short and asserted only the Markdown. The fix goes further than re-clamping: on overflow the behavior now splits by whether the replacement can be matched at all. An **inert** replacement keeps the classic step-back (nothing it could select is ever an insertion — pinned unchanged). A **query-containing** replacement **deselects entirely** (`index = -1`) rather than stepping back, because the "last match before the insertion" can itself be an earlier insertion from the same session — auto-selecting it would re-arm the same habit-driven trap one step removed. Concretely: consuming the last original grays out Replace (a clear "done" signal), the surviving incidental matches stay visible as passive highlights, and the arrows deliberately navigate back into them and re-arm the action. The engine refuses a single replace at `index -1`, and `find()` recovers from that state natively.

Pinned at every layer: the Muya spec now runs the wildcat sequence to exhaustion (asserts `index === -1`, the no-op third replace, `wildwildcat` never appears, and arrow recovery); a controller test covers the refusal before any history boundary is cut; and a new E2E walks `apple -> pineapple` through all three originals — counts `2/3 -> 3/3 -> 0/3`, Replace disabled, Replace All still armed, arrows re-enable.

**Touch targets (P2 — confirmed in substance).** One correction: there is no *documented* 44px contract in the repo — but the substance stands and is worse than stated: at 36px these were the smallest tap targets in the entire app (the floor elsewhere is the 40px top-bar icons; text actions run 42–48px), on adjacent high-frequency controls. All three controls and the replace field now sit at a 44px minimum.

Verification: Muya search specs 15/15, unit suite 606/606, typecheck clean, changed files lint clean, `mobile-editor-search` E2E 13/13. (Process note: the first E2E run of the new tail test failed against a stale reused dev server that still served the pre-fix engine — killed and re-run fresh; the earlier rounds' E2E results were unaffected since inert replacements behave identically on both engines.)

Renakoni:
**Round three verified and fixed in 37d3641** — the finding is correct on all three sub-points, and the arithmetic checks out exactly.

- **Input box**: `.search-input` had no cross-axis rule, so inside the centered flex pill it was one text line tall (16px x 1.2 = 19.2px) — taps in the field's vertical padding hit a non-interactive `<div>`. The input now has `align-self: stretch` on the shared class, which also closes the identical latent gap in the round-one find field (its input had the same 19px box inside a 40px pill since v0.1.0-era styling).
- **Short buttons**: min-height-only left "Aa" at ~41.5px and "All" at ~38.5px wide, exactly as measured. Both controls now carry `min-width: 44px` alongside the height.
- **Regression net**: the expand E2E now asserts the actual `boundingBox()` of the case toggle and both replace actions (>= 44x44) plus the stretched input's height, so the floor is pinned to the interactive elements themselves rather than to a comment.

Also merged current main into the branch (lockfile drift from #172 only, clean merge). Verification: 606 unit tests (vendor-sync contract included), `mobile-editor-search` E2E 13/13 with the new box assertions, typecheck clean.

PR #173, opened 2026-08-05
Add the caret-contextual table panel with seven structure commands

### LINKED ISSUE #144: Structure editing for existing tables
Creating tables works, but there is no way to restructure an existing one from the phone UI.

**Decided design:** when the caret enters a table cell, the bottom toolbar switches to a TABLE panel automatically and restores the previous panel on exit. v1 ships exactly seven commands: insert row above/below, insert column left/right, delete row, delete column, delete table. Column alignment comes later.

**Evidence:** vendored Muya already implements and tests \`Table.insertRow/insertColumn/removeRow/removeColumn\`; the desktop table toolbars document the caret-restore behavior.

**Done when:** a narrow typed adapter drives the seven commands; every command leaves the caret in a surviving cell; undo/redo, first/last row/column, single-row/column tables, and the automatic panel switch are covered by tests and a device check.

Renakoni:
Closes #144.

## Design (as discussed)

The bottom toolbar gains a **caret-contextual TABLE panel**: it exists only while the caret sits in a table cell, switches in automatically on entry, and restores the previously active panel on exit — unless the user manually switched panels while inside, in which case their choice wins. No floating chrome, no sheets, no long-press conflicts: the commands live in thumb reach at exactly the moment they apply.

Seven v1 commands, all relative to the caret cell, each exactly **one undo step** (`history.cutoff()` + `jsonState.flush()` boundary pattern, pinned engine-side since #171):

| Command | Caret afterwards |
|---|---|
| Insert row above / below | First cell of the new row |
| Insert column left / right | The new cell in the CARET's row (resolved via `Table.cellAt`) |
| Delete row / column | Engine-captured surviving neighbour; removing the last one removes the whole table and seats the caret on the outside content |
| Delete table | Outside neighbour resolved at the table boundary before detach |

Decisions from the design discussion: **header-row insert-above stays enabled** with engine semantics (informal tables often have no real header — the rule stays uniform), and **delete-table asks no confirmation** — one-step undo is the safety net; the button renders in `--danger` one tier above row/column deletes.

## Implementation notes

- `tableCommands.ts`: a narrow typed adapter. `resolveTableContext()` climbs content → cell → row → inner → table verifying every block name, and refuses detached subtrees. The mutation is wrapped so the op flush and closing history cutoff run even if the engine throws.
- **Engine fixes that fell out of review** (vendored Muya): a new `Table.outsideContentInContext()` resolves the outside neighbour from the boundary content descendants — the inherited traversal starts at the table's PARENT's siblings and never saw the table's own neighbours; the engine's own `removeRow`/`removeColumn` whole-table fallbacks and the desktop whole-table cut path shared the bug and now use it. A new `ScrollPage.resetToSingleEmptyParagraph()` (the cut path's private recovery, promoted) restores the one-empty-paragraph invariant whenever a removal leaves the document **contentless** — keyed on content descendants, not top-level child count, so a table that was a list item's only child (reachable via the pipe conversion) recovers too.
- The `'table'` panel id lives outside `MOBILE_TOOLBAR_PANELS`, so it can never become the settings default; settings changes made mid-table apply underneath the snapshot.

## Verification

- Vendored-Muya CI gate 1486/1486; new real-tree specs: `outsideContent` 5/5 (top-level next/prev preference, last-row removal onto prose, nested sibling resolution, sole-content contentless predicate + recovery), `resetToSingleEmptyParagraph` 2/2.
- App unit 633/633: adapter 22 (offsets per command, cutoff/flush ordering, survivor/outside/recovered caret, detached refusal, contentless-vs-skeleton distinction), toolbar composable 10 (auto-switch snapshot/restore, manual-choice wins, settings interplay).
- E2E `mobile-table-structure` 7/7: panel follows the caret in/out; inserts/deletes around the caret with typing pinning the caret row (column inserts) and the landing paragraph (removals); table-only documents recover an editable paragraph via delete-table and via successive last-row deletes; the list-item sole-content shape walks the real UI (type `- `, insert from the sheet, delete, type into the recovery). Full `mobile-editor-toolbar` regression 34/34.
- Typecheck and lint clean; ten locales gain the panel + seven command strings.


Renakoni:
**Both findings verified and fixed in 8f9edbe.**

**Column-insert caret (P2 — confirmed by code reading).** `Table.insertColumn()`'s forEach starts at the header row and records only the first cell it creates, so the adapter seated the caret in the header after any body-row insert — and the E2E's count-then-delete sequence was indeed blind to it (deleting the active column passes regardless of which row holds the caret). Fix: the adapter resolves the inserted cell at `(caret row, new column)` through the existing `Table.cellAt()`, with the engine return kept only as a fallback for an unresolvable cell. The E2E now types into the fresh column and asserts the text lands in the body row and NOT in the header; a unit case pins the `cellAt(rowOffset, newColumn)` call and that the header cell's cursor is never touched.

**Empty-page removal (P2 — confirmed, all three routes).** Deleting the document's only table — directly, or through its last row or last column via the engine's null-survivor returns — left a zero-child scroll page: no caret target, and the blank-area click handler dereferences the last child. The fix mirrors the engine's own precedent exactly: the whole-table cut path's private `resetToEmptyParagraph` logic moves onto `ScrollPage.resetToSingleEmptyParagraph()` (the cut path now delegates to it — one truth), and the adapter invokes it **inside the same undo boundary** whenever a removal produces no caret and `scrollPage.length() === 0`, so a single undo restores the table and drops the recovery paragraph. Coverage: a new engine spec for the method (fresh boot + already-emptied page), adapter cases for all three removal commands plus a guard that a non-empty page is left alone, and two table-only-document E2E flows that delete via the table command and via successive last-row deletes, then type into the recovered paragraph.

Verification: engine spec 2/2, unit suite 633/633 (adapter 22), typecheck and lint clean, `mobile-table-structure` E2E 6/6.

Renakoni:
**Round two verified and fixed in bb3bb7b — both findings are real, and the second one is bigger than the adapter.**

**Outside-neighbour traversal (P2 — confirmed; engine-wide).** `nextContentInContext`/`previousContentInContext` are designed for CONTENT blocks: the walk starts at `this.parent`'s siblings, so invoked on the Table block it answers "what follows the table's *parent*" — for a top-level table that's the scroll page, and prose adjacent to the table is never found. Crucially, this wasn't just the adapter's delete-table path: the engine's own `removeRow`/`removeColumn` whole-table fallbacks and the desktop whole-table cut path share the identical call shape. Fix: a new `Table.outsideContentInContext()` resolves the neighbour from the boundary content descendants (`lastContentInDescendant().nextContentInContext()`, else the symmetric previous), whose upward recursion crosses the table edge correctly — and all four call sites now go through it. Pinned on a real tree in a new engine spec: next-preference, previous-fallback when nothing follows, last-row removal seating the caret on the following prose, and the nested-in-list-item shape resolving the item's sibling paragraph. The E2E now **types after both removal paths** and asserts the text lands in the neighbouring paragraph (`landed-after`), closing the caret blind spot exactly as requested. The old fake-table specs were retargeted to the single boundary method (the next-else-previous preference now lives inside it, covered by the real-tree spec).

**Real page vs table.parent (P2 — confirmed, including the throw).** Tables nest inside list items (the createTableImageCursor spec builds exactly that), so casting `table.parent` to ScrollPage was wrong twice over: the emptied list item reports `length() === 0`, and `resetToSingleEmptyParagraph()` doesn't exist on it — TypeError after the table is already detached, skipping the JSON flush and the closing history cutoff. The recovery now resolves the real page via `editor.scrollPage`; an emptied nested container is treated as the non-empty-document case it is — the fixed outside resolution seats the caret on the surrounding content and the empty item stays (the traversal explicitly tolerates empty containers, #4644). The whole mutation is additionally wrapped in try/finally so the op flush and boundary cutoff run regardless of what the engine throws. One documented residue: a pathological document whose only content anywhere is a table inside a single list item no longer throws, but has no outside content to select — the caret stays unset and the empty item remains; out of v1 scope.

Verification: full vendored-Muya CI gate 1486/1486 (the one locally-visible clipboard failure is the pre-existing upstream exclusion), new real-tree spec 4/4, app unit suite 633/633, typecheck and lint clean, `mobile-table-structure` E2E 6/6 with the new typing assertions.

Renakoni:
**Round three verified and fixed in a4b4e20 — the finding is correct, and my previous "pathological shape" call was wrong.** The sole-content nested table is a first-class reachable state: the pipe conversion replaces a list item's paragraph with the table (pinned intentional by listMathEnterConvert), the table panel exposes delete there, and deletion left zero carets — in direct conflict with the issue's done-criterion. Thanks for refusing the hand-wave.

**Fix.** The recovery predicate was measuring the wrong thing: `scrollPage.length()` counts top-level containers, but the invariant that matters is *contentlessness*. A null outside resolution already proves it — the boundary walks cover the entire document in both directions, skipping empty containers — so the recovery now keys on that, with a `firstContentInDescendant()` probe as a guard against future misuse rather than the decision itself. `resetToSingleEmptyParagraph()` clears the leftover skeletons (empty list → item) as part of restoring the invariant, inside the same undo step. The whole-table **cut path had the identical container-counting flaw** and now uses the same content-descendant predicate.

**Coverage, at every layer.** The real-tree engine spec now builds the reachable shape itself — `createTable` from an emptied item paragraph takes the same replace-in-place path as the conversion — and pins the full predicate chain: sole content = a table-cell content as the page's first content descendant; after removal, `length() > 0` while `firstContentInDescendant()` is null; recovery restores exactly one paragraph content. The adapter fake now models content descendants instead of child counts. And a new E2E walks the real UI end to end: type `- `, insert a table from the size sheet (it nests as the item's only child — asserted via `li > figure.mu-table`), delete it from the table panel, and type into the recovered paragraph. (One test-shaping note: programmatic insertion does not emit a selection change, so the panel follows the first tap into a cell — the E2E documents this, and the sheet flow's toolbar collapse is re-expanded as any user would.)

Also refreshed the stale PR body ("Engine untouched" and the old counts) as flagged.

Verification: engine specs — outsideContent 5/5, resetToSingleEmptyParagraph 2/2, table+scrollPage folders 70/70; app unit 633/633; typecheck and lint clean; `mobile-table-structure` E2E 7/7.

Renakoni:
**Round four verified and fixed in a34dd90.** The finding is correct — and it deserves a candid note: I had identified this exact risk during the round-three work and dismissed it as practically unreachable. chunkedMount.spec placing a table at the frontier on purpose settles that; the dismissal was wrong.

**Fix.** The forward walk in `Table.outsideContentInContext()` now uses `resolveNextContentInContext()` — the pending-mount-aware traversal that exists precisely for this (#4887): it keeps extending the mount frontier until content appears or the document genuinely ends. The backward walk stays plain deliberately: progressive mounting is strictly a prefix, so everything before a mounted table is mounted. With that, `null` finally proves what the recovery predicate needs — that the LOGICAL document holds no content outside the table — and `resetToSingleEmptyParagraph()` now documents that precondition explicitly. The adapter and the clipboard whole-table path inherit the correction without changes.

**Regression coverage** (in chunkedMount.spec, where the frontier harness and fake timers live): a table at the frontier with 49 pending paragraphs behind it — `outsideContentInContext()` resolves `Paragraph 551`, and both whole-table removal fallbacks (last-row and last-column) seat the caret there, with `rawState` exact at SECTIONS-1 and `expectDomMatchesState` green after the timers drain, proving the resolution dispatched no stray ops into the removal's undo boundary.

**Self-review pass beyond the finding** (per the review's own standard): a receiver sweep over every remaining `nextContentInContext`/`previousContentInContext` call site in the engine confirms all are CONTENT-block receivers — the four parent-receiver misuses surfaced across this review cycle (mobile delete-table, engine last-row, engine last-column, clipboard whole-table) were the only ones. I also re-walked the adapter's other engine touchpoints for mount sensitivity: inserts, `cellAt`, and row/column survivors all operate strictly inside the (mounted) table; the capture-before-detach ordering holds at all four boundary sites; and the try/finally flush/cutoff pairing is unaffected.

Verification: Muya CI gate 1489/1489 (grew by the frontier regressions), app unit 633/633, typecheck and lint clean, `mobile-table-structure` E2E 7/7.

Renakoni:
**Round five verified and fixed in 2ef93f7.** The finding is correct: the unconditional capture was harmless when the walk was a cheap mounted-tree traversal, but round four's mount-aware resolution turned it into potential synchronous materialization of the next top-level block on EVERY ordinary row delete — exactly the stall progressive mounting exists to avoid, introduced by my own previous fix.

**Fix.** `removeRow` computes the outside neighbour only when no row survives (the whole-table path), still before the detach so the boundary cells are attached for the walk. Caller sweep for the class: `removeRow` was the only conditional-need site — `removeColumn` already resolves only on its single-column path, and delete-table plus the whole-table cut path always need the resolution.

**Regression restructured as prescribed.** The row spec's direct `outsideContentInContext()` assertion did pre-mount the successor and blinded the test; it now pins both behaviors instead: an ordinary row delete leaves `mountedParagraphs` untouched (frontier not advanced), and the last-row removal advances it by exactly one while seating the caret on `Paragraph 551`, with the state count exact and DOM/state equality after the timers drain.

Verification: Muya CI gate 1489/1489, app unit 633/633, typecheck and lint clean, `mobile-table-structure` E2E 7/7.

PR #175, opened 2026-08-05
Record a selection for the session's first undoable op

Renakoni:
Fixes the device-reproduced #144 follow-up: **insert a table row → undo → any further table command silently no-ops and the toolbar panel kicks back** to the previously active one.

> **Delivery contract (accepted in review): this PR is intentionally incomplete on its own — merge #177 immediately after.** #175 fixes the null-selection / detached-block class; recorded **post-op tail paths that disappear on undo** (insert-below on the last row, insert-right on the last column) still break here and are fixed at the root by #177's `primeRecordSelection()`. The explicit tail regressions live in #177 and are expected to FAIL on #175 alone.

## The chain this PR closes

1. The history's selection bookkeeping returns the previous-record stack entry, so the **session's first recorded op always stored `selection = null`**.
2. Undoing that op ran `_restoreSelection(null)`, which restored **nothing** — `editor.activeContentBlock` kept pointing into the subtree the inverse op had just removed (the freshly inserted row's cell, where the insert had seated the caret).
3. The table-context climb walked the detached chain, correctly refused it — the command returned `false` without touching the engine, while the app's caret-state refresh read 'not in a table' and dropped the contextual panel to the pre-table snapshot.

Verified live on the Xiaomi 14 and reproduced byte-for-byte in a happy-dom probe before fixing.

## Fix (scope: the first-record-null class)

- `_getLastSelection()` falls back to the **current** selection when the stack holds a single entry — undoing the first op then re-resolves the recorded path against the post-undo tree. This restores the caret whenever that path still resolves (e.g. insert-above shapes, whose index the original row re-occupies); it is a stack-approximation, not a true before-command selection — hence #177.
- `_restoreSelection(null)` hardening: never leave a **detached** active content block behind (the attachment climb stops AT the scroll page — it has a non-block parent of its own).

## Verification

- New engine specs (history): the adapter sequence through undo to a working follow-up insert, plus the detached-guard case; full Muya CI gate green at this head.
- New E2E replaying the device flow through the real toolbar undo button (insert-above shape): insert → undo → insert again inserts, panel stays TABLE.
- App unit suite, typecheck, and lint clean.

### DIFF
--- a/tests/e2e/mobile-table-structure.spec.ts
+++ b/tests/e2e/mobile-table-structure.spec.ts
@@ -90,6 +90,30 @@ test('inserts and deletes rows and columns around the caret cell', async ({ page
   await expect(page.getByTestId('editor-host')).not.toContainText('fresh')
 })
 
+test('a table command works again right after undoing the previous one', async ({ page }) => {
+  // Device-reproduced #144 bug: insert row -> undo -> insert row again
+  // silently no-opped and kicked the panel out. The session's FIRST
+  // recorded op stored no selection, so its undo restored nothing and
+  // the active content block kept pointing into the removed row.
+  await openTableDraft(page)
+
+  await page.getByTestId('toolbar-expand-button').click()
+  await tapCell(page, 'two')
+  await expect(tableRows(page)).toHaveCount(2)
+
+  await page.getByTestId('toolbar-table-table-insert-row-above').click()
+  await expect(tableRows(page)).toHaveCount(3)
+
+  await page.getByTestId('toolbar-command-edit.undo').click()
+  await expect(tableRows(page)).toHaveCount(2)
+
+  // The caret is back on a live cell: the panel stays TABLE and the
+  // repeated command inserts again instead of no-opping.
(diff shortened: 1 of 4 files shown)


Renakoni:
**Review round two verified — the tail-insertion gap is real, and the fix lives one PR up.** The finding's mechanics check out exactly: the first-record fallback records the POST-op selection, and my regression's insert-above-at-index-1 was indeed the lucky shape whose path the original row re-occupies after undo. Insert-below on the last row / insert-right on the last column park the caret at a tail index the undo removes — the recorded path stops resolving and the context fell to the document head.

Rather than patching the approximation here, the root fix is #177's `primeRecordSelection()` (stamp the entry with the TRUE before-command selection), which subsumes this case for every table command. Per the review's ask, #177 now carries **explicit tail regressions and they pass under priming**: two engine specs (insert-below-last-row and insert-right-last-column undo both return to the origin cell, history folder 20/20) and an E2E repeating the tail flow through the toolbar undo button (`mobile-table-structure` 10/10).

**Merge guidance:** please merge #175 and #177 back-to-back (stacked) — #175 alone leaves the tail class open, as the review correctly concludes.

PR #177, opened 2026-08-05
Restore the caret near its edit after undoing a table command

Renakoni:
Stacked on #175 (same engine region) — **the completing fix for #175's accepted review gap**. Fixes the device-reported caret teleport: **delete the table → undo → the table returns but the caret jumps to the document head**, and closes the disappearing-tail-path class (insert-below on the last row, insert-right on the last column) that #175's stack-approximation cannot cover.

## The chain

The undo entry's recorded selection is the history's one-record-behind approximation — for the delete that meant the **post-delete** caret (the paragraph after the table). Undo re-inserts the table exactly where that recorded path points; the paragraph-shaped path remainder misses inside the table, `queryBlock` resolves nothing, and `_restoreSelection` fell back to `focus()` — the document head. Tail insertions break the same way: the post-op caret sits at a tail index the undo removes, so the recorded path stops resolving entirely.

## Two-layer fix

- **`History.primeRecordSelection()`** — boundary-cut commands stamp the next entry with the TRUE before-command selection right before mutating (the adapter calls it inside the cutoff window, so all seven table commands inherit it). Undoing any table command now puts the caret back exactly where the user was — including the original cell inside the restored table after undoing delete-table, at the exact offset.
- **Longest-resolvable-prefix fallback** in `_restoreSelection` — when a recorded path no longer resolves (or resolves to a container), the caret lands collapsed at the nearest content under the deepest resolvable prefix. Unprimed entries — typing, legacy histories, redo paths — now land local to the edit instead of at the head.

## Verification (head 6f58f45)

- Engine specs pin both layers plus the tail class: primed exactness onto the original cell (`one`); the unprimed fallback landing on the restored table's first content (`A`) — never `before`/head; and **explicit tail regressions** (insert-below-last-row and insert-right-last-column undos returning to the origin cell) that fail on #175 alone. History folder **20/20**; **full Muya gate green** (typing-undo grouping and coalescing semantics untouched).
- E2E replays the device flows through the real toolbar undo button: the delete-table undo with an exact caret-offset assertion (`Xone` typed into the restored cell) and the tail insert-below flow re-inserting after undo. `mobile-table-structure` **10/10**.
- App unit 633/633, typecheck and lint clean. (A pleasing detail from test development: before pinning the caret to the cell start, the restore reproduced the user's original mid-word tap offset exactly — `oXne`.)

### DIFF
--- a/src/features/editor/tableCommands.test.ts
+++ b/src/features/editor/tableCommands.test.ts
@@ -91,7 +91,7 @@ function createFakeEditor({
     offset: vi.fn(),
   }
 
-  const history = { cutoff: vi.fn() }
+  const history = { cutoff: vi.fn(), primeRecordSelection: vi.fn() }
   const jsonState = { flush: vi.fn() }
   const editor = {
     editor: { activeContentBlock: content, history, jsonState, scrollPage },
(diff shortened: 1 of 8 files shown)


Renakoni:
Follow-up to the #175 round-two finding: added the explicit tail-insertion regressions the review asked for, passing under priming (6f58f45) — engine specs for insert-below-on-last-row and insert-right-on-last-column undos returning to the origin cell (history folder 20/20, the shapes whose recorded post-op paths stop resolving entirely), plus an E2E repeating the tail flow through the toolbar undo button (10/10). These would fail on #175 alone, confirming the review's analysis that priming is the completing fix.

PR #178, opened 2026-08-05
Render authored soft breaks as <br> at the export DOM stage

### LINKED ISSUE #143: Soft line breaks are lost in HTML/PDF export

Renakoni:
Fixes #143 (upstream marktext#3676) — second attempt, on a fundamentally different footing than the closed-parked #160.

## Why the first attempt died, and what changed

#160's core was accepted as right: render authored soft breaks as `<br>` on export (CommonMark explicitly permits it), matching what the editor already shows (`.mu-content` is pre-wrap). What died under ten review rounds was the **guard** — deciding "is this newline inside raw HTML?" from marked's token stream requires emulating HTML-parser semantics (raw-text elements, legal nesting, mismatched closes, fragmented tokens, comments, self-closing forms). ~20 of its 28 regression tests existed solely to pin that emulation.

**The fact nobody used: `renderHtml()` already builds a real DOM** (after DOMPurify, for mermaid/diagrams/heading ids). At that stage the BROWSER has authoritatively resolved everything the emulation chased. This PR does the conversion there:

- **The token layer contributes exactly one bit** — the only thing it authoritatively knows: "this newline is markdown text, not raw HTML." A `walkTokens` hook replaces newlines inside marked's `text` tokens with a per-render random sentinel. The marking DIRECTION is the design's security and correctness contract, and it has two legs: **by construction**, escaped text-token content cannot originate tag syntax or comment data, so **raw HTML is never rewritten in any way** — it reaches DOMPurify and the HTML parser byte-identical to a sentinel-free render; **by sink inventory**, the one text-token path whose destination is an ATTRIBUTE rather than a DOM Text node — image labels, which marked renders into `alt` — is explicitly exempted from marking (review caught this: the construction argument alone does not cover destinations, and an unexempted multiline image label leaked the sentinel into `alt`).
- **The DOM pass resolves every sentinel** to either `<br>` (authored soft break in phrasing context under `p`/`li`) or `\n` (protected containers — `pre, code, kbd, script, style, textarea, title, svg, math, table, .katex` — plus structural-container text and block-edge newlines). A total function: nothing can leak into the output, pinned by a sentinel-prefix assertion on every test export. The pass runs before the mermaid/heading-id stages so no other pass ever observes a sentinel.
- **Scope rule:** newlines that arrive as real `\n` — raw HTML formatting, marked's own block separators — are never visited at all. marked's canonical string output is not modified; block separators, `[TOC]` paragraph shape, everything the string protocol's consumers match, untouched by construction.

## Security contract

Review killed two earlier shapes of this feature at the same boundary, and both failures came from rewriting raw HTML:

1. A string-level post-serialization restore let `href="java␤script:alert(1)"` ride through DOMPurify inert and re-materialize as a scheme-splitting newline.
2. A DOM-level attribute restore (newline → space) still re-armed `href="␤javascript:alert(1)"` — URL parsing strips LEADING spaces, so a leading-whitespace downgrade reassembles the scheme; sentinels in tag-syntax whitespace additionally corrupted multiline tags, and comment-data sentinels leaked unrestored.

The flipped marking direction removes the entire attack and corruption surface instead of patching instances: raw HTML is never touched at any stage, so DOMPurify judges the authored bytes and its verdict is final — no pass runs after it that could change what an attribute value means. Multiline start tags and attribute newlines now export byte-verbatim (better fidelity than the previous revisions, not just safer). Raw comment data is never marked either; production DOMPurify removes comments outright, which is pre-existing sentinel-free behavior — the guarantee this feature adds is that no sentinel can enter comment data, pinned via happy-dom where comments stay observable.

Note for reviewers: happy-dom's DOMPurify attribute pass is inert (a probe showed even a literal `javascript:` href survives `sanitize()` there; real Chromium strips it). The suite therefore asserts what this feature owns — byte fidelity of the authored whitespace, i.e. no laundering before DOMPurify or re-arming after — and proves the policy verdict via DOMPurify's string-level `isValidAttribute`, which does work in that environment.

## Parity with the parked ground truths

All of #160's regression ground truths hold **case for case** — including the subtle ones: the comment-adjacent newline rides inside marked's html token and is simply never touched; raw `<p>`/`<blockquote>` formatting whitespace exports byte-identical. The editor and the conformance renderer (`renderToStaticHTML`) are untouched — the pass lives only in `MarkdownToHtml`, which is inherently export-only, so no option flag is even needed.

## Verification

- 35-test suite over the full `renderHtml()`/`generate()` pipeline: 7 positives (paragraph, tight item, inline-wrapped, inside inline spans, blockquote, hard-break intact, final document), structural negatives (block separators, nested/loose lists — multi-child included, code spans, fenced code, tables — nested-in-list included), the raw-HTML war cases (inline `<pre>`/`<textarea>` interiors, mismatched closes, `<pre><code>` nesting), multiline start tags in both newline positions, multiline comment data (a sentinel-leak probe via happy-dom; production DOMPurify removes comments), multiline image labels in inline and reference form (alt stays sentinel-free and byte-identical to marked's own output), verbatim attribute newlines, whitespace-split `javascript:` fidelity + DOMPurify's own verdict on five split variants, sentinel-lookalike authored text, a sentinel-prefix leak assertion on every export, and an 8k-line linearity guard (per-boundary `slice()` copies previously made a long paragraph quadratic — ~3s at 16k lines — now O(1) boundary checks via precomputed emptiness frontiers). Fixtures pad past a happy-dom+DOMPurify first-element quirk (verified absent in real Chromium; documented in the spec).
- Full vendored-Muya CI gate **1530/1530**; app unit 633/633; typecheck and lint clean.
- Android's PDF/share path consumes `generate()` unchanged — the fix arrives there for free.

Groundwork for the export-pipeline blueprint (`ExportResult` + `[TOC]` token) intentionally deferred to its own PR: the DOM-stage design decouples this fix from the string-protocol concerns that made the blueprint's Stage A a prerequisite for the token-level approach.

### DIFF
--- a/third_party/muya/src/state/__tests__/softBreakExportHtml.spec.ts
+++ b/third_party/muya/src/state/__tests__/softBreakExportHtml.spec.ts
@@ -1,0 +1,190 @@
+// @vitest-environment happy-dom
+
+import { describe, expect, it } from 'vitest';
+import { renderToStaticHTML } from '../renderToStaticHTML';
+import { MarkdownToHtml } from '../markdownToHtml';
+
+// marktext#3676 — a soft line break (a bare `\n` inside a block) shows as a
+// line break in the editor (`.mu-content` is pre-wrap) but collapsed to a
+// space on HTML/PDF export.
+//
+// Second attempt (#160/#4951 post-mortem): the token-level guard — deciding
+// "is this newline raw HTML?" from marked's token stream — required
+// emulating HTML-parser semantics and died under review. This design moves
+// the conversion to the DOM stage `renderHtml` already owns: after
+// sanitize + innerHTML, the BROWSER has authoritatively resolved raw-text
+// elements, nesting, and mismatched tags, so one `closest()` covers the
+// entire protection surface. The token layer contributes the single bit
+// only it knows — "this text came from a raw HTML token" — as a newline
+// sentinel, with no tag tracking at all.
+//
+// The parked suite's ground truths all hold: the sentinel carries every
+// raw-token newline (comment-adjacent ones included — marked folds them
+// into the html token) through the DOM pass verbatim, so the output
+// matches the token-level design case for case, minus the state machine.
+
+// happy-dom quirk: DOMPurify under happy-dom strips the FIRST element's
+// wrapper tags (verified directly: `<p>a</p>` -> `a`; real Chromium does
+// not do this). Every fixture leads with a sacrificial pad block so the
+// content under test never sits first. Test-environment workaround only.
+const PAD = '# pad\n\n';
+
+async function exportHtml(markdown: string) {
+    return new MarkdownToHtml(PAD + markdown).renderHtml();
+}
+
+describe('marktext#3676 — authored soft breaks render as <br> on export', () => {
+    it('renders a paragraph soft break as <br>', async () => {
+        expect(await exportHtml('line one\nline two\n')).toContain(
+            '<p>line one<br>line two</p>',
+        );
+    });
+
+    it('renders a tight-item soft break as <br>', async () => {
+        expect(await exportHtml('- line A\n  line B\n')).toContain(
+            '<li>line A<br>line B</li>',
+        );
+    });
+
+    it('renders a soft break between inline-wrapped lines as <br>', async () => {
+        expect(await exportHtml('- **left**\n  **right**\n')).toContain(
+            '<strong>left</strong><br><strong>right</strong>',
+        );
+    });
+
+    it('renders a soft break inside an inline span as <br>', async () => {
+        expect(await exportHtml('**left\nright**\n')).toContain(
+            '<strong>left<br>right</strong>',
+        );
+    });
+
+    it('renders a blockquote paragraph soft break as <br>', async () => {
+        expect(await exportHtml('> line one\n> line two\n')).toContain(
+            'line one<br>line two',
+        );
+    });
+
+    it('keeps a real hard break working (two trailing spaces)', async () => {
+        expect(await exportHtml('line one  \nline two\n')).toMatch(
+            /line one<br\s*\/?>\s*line two/,
+        );
+    });
+
+    it('reaches the final document through generate()', async () => {
+        const doc = await new MarkdownToHtml(PAD + '- line A\n  line B\n').generate({
+            title: 't',
+            inlineStyles: false,
+        });
+        expect(doc).toContain('line A<br>line B');
+    });
+
+    it('leaves the conformance renderer untouched', () => {
+        const html = renderToStaticHTML('line one\nline two\n');
+        expect(html).toContain('line one\nline two');
+        expect(html).not.toContain('<br>');
+    });
+});
+
+describe('everything that is NOT an authored soft break stays untouched', () => {
+    it('keeps canonical block separators between paragraphs', async () => {
+        const html = await exportHtml('para one\n\npara two\n');
+        expect(html).toContain('</p>\n<p>');
+        expect(html).not.toContain('<br>');
+    });
+
+    it('keeps the newline before a nested list structural', async () => {
+        const html = await exportHtml('- a\n  - b\n');
+        expect(html).not.toContain('a<br>');
+        expect(html).toContain('<ul>');
+    });
+
+    it('preserves an authored &nbsp; item next to a nested list', async () => {
+        const html = await exportHtml('- &nbsp;\n  - child\n');
+        expect(html).not.toContain('<br>');
+    });
+
+    it('keeps loose-item paragraph boundaries structural', async () => {
+        const html = await exportHtml('- para one\n\n  para two\n');
+        expect(html).not.toContain('<br>');
+    });
+
+    it('collapses a code-span newline to a space per spec, never <br>', async () => {
+        expect(await exportHtml('a `x\ny` b\n')).toContain('<code>x y</code>');
+    });
+
+    it('keeps fenced code content verbatim', async () => {
+        const html = await exportHtml('```\nline one\nline two\n```\n');
+        expect(html).toContain('line one\nline two');
+        expect(html).not.toContain('line one<br>');
+    });
+
+    it('keeps raw-HTML formatting newlines verbatim (raw <p>)', async () => {
+        // At the DOM stage a raw-authored <p> and a markdown paragraph are
+        // the same element — the raw-newline sentinel is what keeps this
+        // distinction alive across the parse.
+        const html = await exportHtml(
+            '<p>\n<strong>left</strong>\n<strong>right</strong>\n</p>\n',
+        );
+        expect(html).toContain(
+            '<p>\n<strong>left</strong>\n<strong>right</strong>\n</p>',
+        );
+        expect(html).not.toContain('<br>');
+    });
+
+    it('keeps raw blockquote inline whitespace untouched', async () => {
+        const raw
+            = '<blockquote><strong>left</strong> <strong>right</strong></blockquote>';
+        expect(await exportHtml(`${raw}\n`)).toContain(raw);
+    });
+
+    it('keeps inline raw-text element content verbatim (<pre>)', async () => {
+        // The war case of the token-level attempt: marked hands the interior
+        // of an inline <pre> over as ordinary text tokens. Post-parse, the
+        // DOM knows exactly where the pre is.
+        const html = await exportHtml('text <pre>a\nb</pre> more\n');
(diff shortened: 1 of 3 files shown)


Renakoni:
Review round 1 addressed in 12b6e3c — all three findings were valid; the P1 in particular was a real sanitizer bypass, and fixing it properly meant moving the restoration stage, not patching the pattern.

**P1 (security, blocking) — sentinel restoration after the sanitization boundary.** Confirmed exploitable as described: `href="java␤script:alert(1)"` sailed through DOMPurify as an inert sentinel-carrying value, and the post-serialization string swap re-created the newline that URL parsing strips — a working `javascript:` URI in the exported document. The restore now runs **on the sanitized DOM, before `innerHTML` serialization**, so no textual mutation ever follows sanitization:
- Text-node sentinels restore the real `\n` — text mutation cannot cross into markup by construction.
- Attribute sentinels restore a **space** rather than `\n`. Space is the provably safe direction: URL parsing strips tab/newline/CR but preserves interior spaces, so a scheme can never reassemble around it — whereas *deleting* the sentinel would concatenate `java`+`script:`. Attribute newlines are collapsible whitespace semantically, so the space is presentation-equivalent. Sentinels that landed in attribute-name position (a newline between attributes) parse into junk attributes DOMPurify already strips.
- New test pins the exact attack string: no `javascript:`, no `java␤script`, no entity-encoded variant in the output.

**P2 (structural) — `<br>` leaking between list items / table sections.** Fixed at two levels: direct text children of structural containers (`UL`/`OL`/`MENU`/`DL`/`TABLE`/`THEAD`/`TBODY`/`TFOOT`/`TR`/`COLGROUP`/…) are now categorically serializer whitespace — that's the rule the multi-child nested list needed, since the inner `<ul>`'s inter-`<li>` text nodes pass a `closest('p, li')` check via the *outer* item. `LI` and the table-section tags also joined `BLOCKISH_TAG` for the edge-adjacency path, and `table` joined the protected selector outright (GFM cells can't contain authored newlines, so there is nothing to convert under a table). New tests: `- parent / - c1 / - c2` and a table nested in a list item, both `<br>`-free.

**P3 (correctness) — fixed sentinel collides with authored content.** The sentinel is now freshly **random per render** (16 crypto bytes, hex, plain ASCII). This closes both the literal collision and the vector an input-scan check could never close: authored numeric character references (`&#xE000;`-style) that only *become* the sentinel character after marked entity-decodes them. With a per-render random token there is nothing stable to author against. New test pins a sentinel-lookalike literal surviving verbatim.

Verification: soft-break suite 27/27 (23 prior + 4 new), full Muya gate 1522/1522, app unit 633/633, typecheck clean. PR description updated to match the current design.

Renakoni:
Round 2 addressed in 28e1fc5. All four findings verified as real — and findings 1–3 share one root cause, so the fix is a design change rather than three patches.

**Root cause of P1×2 + P2(comments):** the sentinel lived INSIDE raw HTML tokens — a textual rewrite of markup. Anything that rewrites markup text must then reason about tag syntax, attribute values, attribute names, and comment data separately, and every reviewed revision missed at least one of those surfaces (leading-whitespace URL stripping this round; interior whitespace the round before; tag-syntax gluing and comment leakage as collateral).

**The fix flips the marking direction.** The sentinel now rides inside marked's `text` tokens — escaped markdown text by construction, which can never sit in tag syntax, an attribute, or a comment. Raw HTML is never rewritten at any stage:

- *Finding 1 (leading-newline `javascript:` bypass):* DOMPurify now judges the authored bytes — `href="␤javascript:alert(1)"` reaches it with the real newline, its URI policy strips attribute whitespace before testing, verdict final. No pass of ours runs after sanitization that could change what an attribute value means, because no pass of ours touches attributes at all. One environment note: happy-dom's DOMPurify attribute pass turned out to be inert (a probe showed a literal `javascript:` href surviving `sanitize()` there, while `isValidAttribute` correctly rejects it and real Chromium strips it). The tests therefore pin the property this feature owns — byte fidelity of authored whitespace through the whole pipeline, i.e. no laundering — plus DOMPurify's string-level verdict on five whitespace-split variants (leading/interior/trailing newline, leading space, interior tab).
- *Finding 2 (multiline tags corrupted):* raw tag-syntax whitespace is never replaced, so `<div id="a"␤class="kept">` and `<div␤class="kept">` parse exactly as authored. New tests assert the attributes exist in the output, not just substring survival.
- *Finding 3 (comment sentinels never restored):* raw comment data is never marked, so nothing needs restoring — `before <!-- left␤right --> after` exports byte-verbatim. New test pins it, and the suite now asserts the sentinel prefix appears in NO export output, closing the leak class rather than the instance.
- *Finding 4 (quadratic slices):* per-boundary `slice()` prefix/suffix copies replaced with two precomputed emptiness frontiers; each boundary check is O(1). An 8k-line single-paragraph test guards linearity (the whole 33-test suite runs in ~650ms including it; the quadratic version took ~740ms for that input alone).

Net effect on fidelity, beyond safety: attribute newlines, multiline start tags, and multiline comments now export byte-identical to the authored raw HTML — the two previous revisions degraded or corrupted all three.

Verification: soft-break suite 33/33, full Muya gate 1528/1528, app unit 633/633, typecheck clean. PR description rewritten to match the flipped design.

Renakoni:
Round 3 addressed in 3e8ae02.

**Image-alt sentinel leak (P2) — confirmed and fixed at the invariant level.** The finding is exactly right: my "text tokens are escaped text by construction" claim conflated *provenance* with *destination*. The provenance half holds (text tokens can never originate tag syntax), but marked's image renderer sends label text tokens to an ATTRIBUTE destination, where the SHOW_TEXT pass can never resolve them. The fix restores the real invariant — only text tokens that will land in DOM Text nodes get marked: `walkTokens` visits parents before children, so the hook tags the image token's entire label subtree in a WeakSet at the image visit and skips those text tokens. Nested structures (`*a\nb*`) are covered by the recursive tagging, reference images by sharing the same token type, and multiline alt keeps marked's own newline behavior byte-for-byte. No post-sanitization attribute mutation is reintroduced. I also swept marked's core renderers for other text-token→attribute sinks: image alt is the only one (link labels render into element content; link/image titles come from `token.title`, which the hook never touches).

Both requested regressions added — inline `alt\ntext` and reference `![alt\ntext][id]` — asserting `alt="alt\ntext"` matches the sentinel-free marked output; the leak itself is covered by the suite-wide sentinel-prefix assertion that every `exportHtml` call already runs.

**Comment-fidelity wording — adopted as suggested.** The test is renamed and re-commented to claim what it actually proves: no sentinel ever enters comment data, with happy-dom keeping comments observable as the leak probe; production DOMPurify removes comments outright, which is pre-existing sentinel-free behavior. The PR description's "byte-verbatim comment" phrasing is corrected the same way.

Verification: soft-break suite 35/35, full Muya gate 1530/1530, typecheck clean.

Renakoni:
The round-4 documentation finding is addressed in 78b4fda plus a PR-body sync.

You are right that the stale "escaped text by construction, so the sentinel can never sit in … an attribute" sentence was exactly the reasoning the image-alt finding disproved — and that leaving it in the interface doc invited a future maintainer to delete the image exemption on the strength of a false invariant. The comments (interface doc, module header, spec header) and the PR body now state the guarantee as two distinct legs: BY CONSTRUCTION, text-token content cannot originate tag syntax or comment data (raw HTML flows byte-identical to a sentinel-free render); BY SINK INVENTORY, a text token's destination is not guaranteed to be a Text node — image labels render into `alt`, the one such sink in marked's core renderers, and are explicitly exempted, with the doc warning that removing the exemption re-leaks the sentinel. The verification numbers in the body are also synced to the actual head (35-test suite, Muya gate 1530/1530), and the multiline-image and comment-probe cases are now listed in the coverage enumeration.

Comment/description-only round; suite re-run 35/35, typecheck clean.

Issue #153, opened 2026-07-16
Bound persisted SAF URI grants with a conservative lifecycle


Renakoni:
The native plugin calls \`takePersistableUriPermission\` and never releases grants. Long-lived use will approach provider/OS grant limits.

**Done when:** a conservative LRU tied to protected/recent documents is defined; a grant is released only when provably no protected document depends on it; a many-documents stress test does not exhaust the provider/OS grant quota.

PR #179, opened 2026-08-05
Make every draft reachable past the home recency cap

### LINKED ISSUE #151: Make every draft reachable beyond the home projection cap
The persistence layer no longer destructively caps drafts, but the home screen's recent projection is bounded — old drafts can exist with no UI path to them.

**Direction:** keep the quiet recent-first home and add an "all drafts" surface; add search/pagination only if measurements demand it. No storage backend migration.

**Done when:** with 25+ drafts mixed with Android recents, every draft can be found and opened without deleting other documents; restart, ordering, and paging never modify the durable collection.

Renakoni:
Closes #151.

## The problem

Drafts are stored completely — nothing is ever destructively capped — but the home list is a **recency projection with a hard cap**. A draft pushed past the cap kept existing with no UI path to it: to the user, indistinguishable from data loss.

## The design (as discussed)

- The resting home keeps its quiet, capped recent-first view.
- A **'Show N more'** text action closes the Earlier section whenever older documents are hidden. It expands the list **in place** to the complete collection, so every existing row affordance — open, long-press multi-select, pin, delete, rename — works on deep items with zero new machinery.
- Expansion is **ephemeral presentation state**: it never writes storage (pinned in the E2E byte-for-byte).
- **Pinned documents beyond the cap stay visible in the resting view** — pinning is explicit intent and must not silently un-surface when newer items push the pin past the cap.
- The cap rises **50 → 100** (review discussion): double Word's own recent-list ceiling (0–50), with headroom under Android ≤ 10's 128-persisted-grant system limit for Android-document records carrying SAF references. Their fine-grained grant lifecycle remains #153; local drafts are never bounded by this number at all.

## Hardening after review

- **The cap limits computation, not just rendered rows.** The first cut materialized list items (whose statistics scan each draft's full Markdown) for the complete collection even while collapsed, then filtered back — and a selection-retention watcher kept that work hot on every autosave. The projection is now layered with an honest cost model: list-item STATISTICS (full-Markdown scans per draft) are materialized only for the records the current state actually renders; the records level is cheaper but not free (local-draft records derive their title from the draft's Markdown), so the selection-retention watcher short-circuits its source and unsubscribes from the projection entirely while no selection is active — a watcher must evaluate its source on every dependency change, so a callback-side guard could not prevent autosaves from pulling the chain hot off-home (verified with a reactivity probe: guard off, zero projection evaluations on a draft change).
- **The Android recents storage cap is pin-aware.** That store is an Android document's only durable home, and the cap was pin-blind at all three layers (upsert, serialize, parse): with 100+ records, opening one more evicted the oldest record even if pinned, and startup pruning then deleted the orphaned pin permanently. Writes now carry the pinned id set and evict the oldest UNPINNED records; reads are uncapped (the cap is enforced pin-aware on every write, and re-capping on read without pin knowledge would undo it). The limit is a TOTAL: protected records claim slots with priority and the newest unpinned records fill the remainder, so the store never exceeds 100 (pins are separately capped at 50, so protection alone cannot overflow it) and the 128-grant headroom argument stays intact. Rename ordering is part of the contract: a SAF rename can change the record id, so the pin migrates to the new id BEFORE the recents write — persisting first with the stale pin set would evict a cap-exempt pinned record on that very write.
- **Expansion keeps focus and announces itself.** The show-all button removes itself once nothing is hidden, which dropped keyboard/switch-access focus to `<body>`; the home now snapshots visible ids at activation, moves focus to the first newly revealed row in DOM order (revealed records interleave under title/ascending sorts), and reports the revealed count through an always-mounted polite live region.

## Verification

- Unit: resting projection caps at 100 while an explicit limit reads the complete collection; records-level sorted projection is uncapped and stats-free; item materialization preserves pre-sorted order; pin-protected cap crossings at the upsert, serialize/parse, and storage round-trip layers (100 unpinned + pinned-oldest + one more ⇒ oldest unpinned evicted, pinned record survives reload).
- E2E (105 seeded drafts): resting view shows exactly 100 with the expander naming the hidden 5; expansion reaches all 105 with the drafts store **byte-identical** before/after; the previously unreachable oldest draft opens and renders; keyboard activation lands focus on the first newly revealed row with the polite announcement present; under title-ascending sort with a single hidden record, focus lands on the Continue masthead that record becomes; pinning a beyond-cap draft in the expanded view keeps it visible after a fresh visit in the resting view.
- Home regression: document-management, navigation, and documents-settings suites green; 641 unit tests, typecheck and lint clean; ten locales gain the expander and announcement strings. The rename cap-crossing regression was verified in BOTH directions (temporarily reverting the pin-migration order makes it fail; the shipped order passes) and extends through the restart chain (startup pin pruning keeps the migrated pin).

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -194,6 +194,7 @@ import { AUTO_APP_LOCALE, translateKnownText, useI18n } from './lib/i18n'
 import {
   createRecentDocumentFromLocalDraft,
   getRecentDocumentListItems,
+  normalizeRecentDocuments,
   type RecentDocumentRecord,
 } from './lib/recentDocuments'
 
@@ -398,10 +399,36 @@ const recentDocumentRecords = computed(() =>
   ],
 )
 const recentActivityDocumentItems = computed(() => getRecentDocumentListItems(recentDocumentRecords.value))
-const documentItems = computed(() =>
-  getSortedRecentDocumentListItems(recentDocumentRecords.value, documentSettings.value),
-)
+// The home's "show all" expansion (#151): the resting view keeps the quiet
+// recency cap, the expanded view reads the COMPLETE collection. Ephemeral —
+// never persisted, so paging can never modify the durable stores.
 const pinnedDocumentIds = computed(() => getPinnedDocumentIds(pinnedDocuments.value))
+const homeListExpanded = ref(false)
+const allDocumentItems = computed(() =>
+  getSortedRecentDocumentListItems(
+    recentDocumentRecords.value,
+    documentSettings.value,
+    Number.POSITIVE_INFINITY,
+  ),
+)
+const documentItems = computed(() => {
+  if (homeListExpanded.value) {
+    return allDocumentItems.value
+  }
+
+  // The resting view = the recency-capped set, plus any PINNED document
+  // beyond the cap — pinning is explicit intent and must keep an item
+  // visible without expanding.
+  const cappedIds = new Set(
+    normalizeRecentDocuments(recentDocumentRecords.value).map(record => record.id),
+  )
+  return allDocumentItems.value.filter(
+    item => cappedIds.has(item.id) || pinnedDocumentIds.value.has(item.id),
+  )
+})
+const hiddenHomeDocumentCount = computed(
+  () => allDocumentItems.value.length - documentItems.value.length,
+)
 const homeDocumentSections = computed(() =>
   partitionHomeDocumentItems(documentItems.value, pinnedDocumentIds.value),
 )
@@ -2053,6 +2080,7 @@ onBeforeUnmount(() => {
     :continue-document="continueDocument"
     :pinned-documents="pinnedHomeDocuments"
     :earlier-documents="earlierDocuments"
+    :hidden-document-count="hiddenHomeDocumentCount"
     :notice="displayHomeNotice"
     :selection-active="homeSelection.isActive.value"
     :selection-count="homeSelection.count.value"
@@ -2064,6 +2092,7 @@ onBeforeUnmount(() => {
     @new-document="newDocument"
     @open-document="openDocument"
     @open-file="openFileFromAndroid"
+    @show-all-documents="homeListExpanded = true"
     @set-tab="setHomeTab"
     @set-settings-page="setSettingsPage"
     @select-document="homeSelection.beginWith"
--- a/src/App.vue
+++ b/src/App.vue
@@ -86,7 +86,7 @@ import {
 } from './features/document-session/documentSessionState'
 import {
   getDocumentSettings,
-  getSortedRecentDocumentListItems,
+  getSortedRecentDocumentRecords,
 } from './features/document-session/documentSettings'
 import { createAutosaveScheduler } from './features/document-session/autosaveScheduler'
 import { createCurrentDocumentPersistence } from './features/document-session/currentDocumentPersistence'
@@ -195,6 +195,7 @@ import {
   createRecentDocumentFromLocalDraft,
   getRecentDocumentListItems,
   normalizeRecentDocuments,
+  toRecentDocumentListItems,
   type RecentDocumentRecord,
 } from './lib/recentDocuments'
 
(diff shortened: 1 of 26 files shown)


Renakoni:
Both review rounds addressed in 2bb05d0 — the eager-stats finding from this round plus the two findings from the earlier round (pin-blind storage cap, focus loss on expansion), all three verified against the code before changing anything.

**Collapsed projection now computationally capped.** Confirmed exactly as described: `allDocumentItems` (Infinity) materialized statistics for every draft's full Markdown before the collapsed filter, and `watch(documentItems, ...)` kept that computed hot on autosave updates even off the home screen. The projection is now layered — `sortedRecentRecords` (records-level, timestamp/title sort keys only) backs ordering, cap membership, and `hiddenHomeDocumentCount`; `documentItems` materializes stats only for the visible records (capped ∪ pinned when collapsed, everything only after explicit expansion); the selection-retention watcher observes the records level so the items computed stays lazy. Unit tests pin the records projection (uncapped, stats-free) and the order-preserving item materialization.

**Pinned Android records protected from the storage cap** (round-1 P2). The eviction chain was real at all three layers — `upsertRecentDocument`, `serializeRecentDocuments`, and `parseRecentDocuments` all applied `slice`-by-recency with no pin knowledge, and the startup orphan-prune (App.vue) then made the pin loss permanent. Writes now thread the pinned id set through `rememberAndroidRecentDocument` / `markSavedAndroidRecentDocument` / `writeStoredAndroidRecentDocuments` and evict the oldest UNPINNED record; the read side is uncapped, since the cap is enforced pin-aware on every write and re-capping on read (which has no pin set) would undo it. The requested cap-crossing regressions exist at three layers: upsert (100 unpinned + pinned oldest + one more ⇒ oldest unpinned evicted, pin survives), serialize→parse round-trip, and the storage read/write pair.

**Focus and announcement on expansion** (round-1 P2). The self-removing button now snapshots the visible ids at activation; after the list grows, focus moves to the first newly revealed row in DOM order (correct under title/ascending sorts where revealed records interleave rather than append), and an always-mounted `role="status"` region announces the revealed count (mounted before it is populated, so polite live regions actually fire). New keyboard E2E activates the expander with Enter and asserts both the resulting `document.activeElement` and the announcement; ten locales gain the announcement string.

Verification: focused suites (recentDocuments 18, documentStorage, documentSettings, androidRecentDocuments, locale contract) green; full app unit 639/639; show-all E2E 3/3 including the new keyboard case; typecheck and lint clean. PR description updated.

Renakoni:
Round 3 addressed in 716e59a. All three findings verified against the code and fixed; the P1 in particular is exactly the data-loss chain the pin-aware cap was built to prevent, reopened through an ordering hole I introduced.

**P1 — pin migration ordered before the recents write.** Confirmed step by step: `renameRecentDocumentWorkflow` builds `updatedRecord` as `{ ...record, ...renamed fields }`, so it keeps the old timestamps; the rename flow persisted the recents while `pinnedDocumentIds` still contained only `previousId`; the pin-aware serialization therefore saw the renamed id as an ordinary oldest unpinned record and evicted it, and the pin then migrated onto a record that startup pruning deletes. The delayed observability you described is also real — the storage helper returns the untrimmed input, so the session looked fine until reload. The migration block now runs BEFORE `persistAndroidRecentDocuments` (the persist callback reads the live pin set, so the write sees the migrated id). The requested regression drives the full chain at the actions layer: 100 newer unpinned + pinned oldest, rename with URI change, and a persist mock that mirrors the App wiring by applying the pin-aware cap with the pin set *as it is at call time* — renamed id present after the storage round-trip, old id gone, pin migrated. (Without the reorder this test fails: the renamed record is 1-over-cap and unprotected at write time.)

**P2 — the cap is now a true total.** Your arithmetic was right: protected records not counting against the limit meant "100 unpinned + up to 50 pinned = 150", contradicting both the constant's own comment and the 128-grant headroom argument. Semantics A adopted as recommended: protected records claim slots with priority, the newest unprotected records fill `limit - protectedCount`, and the store never exceeds 100 (pins are separately capped at 50, so protection alone cannot overflow the total). The 150-record worst case is pinned in a test asserting the round-trip holds exactly 100 with all 50 pins retained, and the constant's documentation now states the total-cap semantics explicitly.

**P2 — Continue masthead joins the focus scan.** Confirmed via `partitionHomeDocumentItems`: the Continue card is simply `items[0]` of the user-sorted list, so under title/ascending sort the only revealed record becomes the masthead and the old `[data-doc-id]` scan had nothing to focus. The Continue button now carries the same `data-doc-id` marker (one shared attribute, per your drift concern), and a new E2E covers the exact scenario — 101 drafts, title ascending, single hidden record titled to sort first — asserting the active element is the Continue card and the live region announces 1.

Verification: focused units 39/39 (recentDocuments / documentStorage / homeDocumentActions), full app unit 641/641, show-all E2E 4/4, typecheck clean.

Renakoni:
Round 4 addressed in 200fcab. Both findings were correct, and finding 2 caught something worse than a test bug — a verification claim I made without running it.

**Records projection pulled hot by the watcher (P2).** Mechanism confirmed: a watcher must evaluate its source on every dependency change to detect one, so the callback-side guard was irrelevant — every draft autosave re-ran per-draft title derivation over full Markdown bodies plus two normalizations and the settings sort, selection active or not. Fixed with exactly the suggested shape: the source getter short-circuits before touching `visibleRecentRecords` when no selection is active, so Vue's dynamic dependency tracking drops the subscription entirely. Verified with a reactivity probe this time (guard off: zero projection evaluations on a draft change; guard on: subscription resumes). The projection comment and the PR description now state the honest cost model — statistics are capped to rendered records; title derivation belongs to the records layer and runs only while something actually renders it. The deeper option (cheap timestamp-only membership for modified/created sorts, cached titles for title sort) is real but out of this PR's scope; with the unsubscribe, the remaining cost occurs only while the home list is actually displayed, which is render work the list needs anyway.

**Rename regression fixture inverted (P2).** Exactly right, and the part that stings: I claimed "without the reorder this test fails" without running that counterfactual — the shared fixture's 2026-07-09 timestamps made the rename target the NEWEST record in the set, so the test was green on the pre-fix ordering and pinned nothing. The target's timestamps are now explicitly below the fleet (with a comment recording the inversion so it cannot silently regress), and the discriminating power was verified in both directions for real: with the pin migration temporarily moved back after the recents write, the test fails; with the shipped ordering it passes. The assertion also now extends through the restart chain — `prunePinnedDocuments` against the stored ids keeps the migrated pin — rather than stopping at the in-memory ref.

**PR description (P3)** updated alongside: 641 unit tests, the Continue-masthead E2E case, and the records-layer claim corrected to the two-tier cost model above.

Verification: homeDocumentActions 11/11 (including the now-discriminating regression, failure direction exercised), full app unit 641/641, show-all + document-management E2E 7/7, typecheck clean.

Issue #180, opened 2026-08-06
Source code mode: edit raw Markdown in a plain-text view


Renakoni:
Spun out of #149: `sourceCodeModeEnabled` persists a value but no source code mode exists. Rather than hiding the control, implement the mode — the WYSIWYG editor is a poor fit for precision edits on mobile (table syntax, nested-list indentation, constructs the engine normalizes), and a raw-text view is the standard escape hatch.

## Why this is cheaper than it looks

The vendored Muya core already ships the desktop source-mode hand-off surface:

- bulk `setMarkdown` designed to coalesce a whole source-mode editing session into ONE undo step;
- WYSIWYG caret restoration from a source-mode `{ line, ch }` position.

Both paths were exercised and hardened by the recent undo/caret PRs (#175, #177).

## MVP scope

- **View**: a monospace `<textarea>` holding the raw Markdown — NOT CodeMirror. A native textarea gets correct mobile IME behavior for free (hard-won lesson from the MIUI keyboard work) at zero dependency cost.
- **Entry**: a toggle in the editor menu; the editor toolbar strips hide while the mode is active.
- **Hand-off in**: current Markdown snapshot → textarea, caret placed near the WYSIWYG caret.
- **Hand-off out**: textarea content → `setMarkdown` as a single undo step, caret mapped back via the existing `{ line, ch }` API.
- **Persistence**: the document state syncs from the textarea (debounced) so autosave, recovery drafts, share, and export keep working while the mode is active.
- **Setting**: `sourceCodeModeEnabled` wires up as "open documents in source code mode" — the stored-only row from #149 becomes a real runtime setting.

## Out of scope (follow-ups if wanted)

Syntax highlighting, find/replace inside the raw view, line numbers.

**Done when** the mode round-trips content losslessly with one undo step per session, the caret survives both hand-offs, autosave/recovery work while active, and the settings governance test counts `sourceCodeModeEnabled` as `runtime`.

Renakoni:
Implementation opened as #182 (engine hand-off audit spec + full shell integration + emulator walkthrough). #181 removes the companion stored-only row; the two PRs together reach #149's done state — the settings governance stored-only list is empty.

Renakoni:
[...]

Issue #149, opened 2026-07-16
Wire or hide the two stored-only settings


Renakoni:
\`sourceCodeModeEnabled\` and \`preferHeadingStyle\` persist values but change no runtime behavior — the controls look functional and are not.

Either implement the real behavior or hide/remove the rows. **Done when** the settings governance test shows no user-facing stored-only interactive rows.

Renakoni:
Resolution split per discussion: #181 removes the placebo `preferHeadingStyle` row; #182 implements source code mode (#180) and wires `sourceCodeModeEnabled` as a real runtime setting. With both merged, the governance test's stored-only list is empty — this issue's done state.

Renakoni:
Done state reached — the settings governance stored-only list is empty, split across two PRs as discussed: #181 removed the placebo `preferHeadingStyle` chooser (Muya's heading command writes ATX unconditionally; Setext covers only h1/h2, near-zero value to wire), and #182 implemented source code mode (#180) and wired `sourceCodeModeEnabled` as a real runtime setting (open documents in source mode). Every remaining interactive settings row is either `runtime` or explicitly `unfinished`-listed, enforced by the governance test.

PR #182, opened 2026-08-06
Add source code mode: raw Markdown editing in a plain textarea

Renakoni:
Implements #180. Together with #181 this completes #149: `sourceCodeModeEnabled` flips from `storedOnly` to `runtime` and the governance list's stored-only set is now empty.

## Design

A toggle in the editor action sheet swaps the WYSIWYG surface for a monospace `<textarea>` holding the raw Markdown. Deliberate choices:

- **Plain textarea, not CodeMirror**: native IME behavior (hard-won lesson from the MIUI keyboard work) and native in-mode undo at zero dependency cost.
- **No syntax highlighting**: the mode's whole point is the unrendered source; highlighting adds dependencies and IME risk for negative value.
- **Theme adaptation, yes**: colors ride the same CSS variables as the editor shell (dark themes included), same card geometry so the toggle doesn't jump. Pinned by an E2E asserting the textarea background equals the `--surface` token.

## Engine audit first (nothing trusted)

The hand-off composes three Muya APIs. Before building the shell, a new engine spec (`sourceModeHandoff.spec.ts`, 6 cases) pinned the exact composition the app runs — and surfaced two ground truths that shaped the design:

1. **`getSelection()`/`getCursorOffset()` read the live DOM selection only** — cleared ranges mean null. Entry state is therefore captured at toggle time, and every capture tolerates null with deterministic fallbacks (caret at document start, explicit null `recordSelection`).
2. **The exit hand-back re-serializes to Muya's canonical markdown** (table columns padded, etc.) exactly as if the content had been typed in WYSIWYG — semantic fidelity, not byte fidelity, is the contract, and the spec asserts against a reference render rather than the raw input.

The composed invariants pinned: `replaceContent(text, entrySelection)` is ONE undo boundary that survives `setCursorByOffset`'s internal double rebuild; the first undo after exit restores the pre-session document AND the entry caret; an emptied textarea exits into the single-empty-paragraph invariant; an unchanged exit records no boundary. Two review rounds hardened the caret hand-back into an explicit contract: the exit coordinates are mapped through the RAW textarea text (the serialization they are expressed against), and a caret parked between syntax characters — after a table row's closing pipe, inside a delimiter row — walks backward one character at a time, each candidate validated by the parse itself, until it resolves: **the caret restores to the nearest editable position at or before where the user left it**, with a deterministic document-start fallback (never a silent loss). Both directions are executed in the regressions — the raw-text mapping and the boundary snap each assert that the OLD behavior provably fails on the same inputs.

## The mode owns the document while active

An `editorSession` markdown override at the single snapshot choke point makes the textarea the document for EVERY consumer — autosave, share, PDF export, save-to-device, and lifecycle flushes — and textarea input drives the same `syncMarkdown` pipeline as WYSIWYG edits (status, autosave scheduling, search/resume invalidation). Surfaces that act on the muya tree stand down: markdown toolbar, search, outline, selection toolbar, link overlay, resume card. Android Back dismisses the mode like a panel (unit-pinned priority: overlays first, then the mode, then leave); editor teardown resets the mode without a hand-back (the document state already tracks the textarea). The wired setting opens documents in source mode **without popping the keyboard**, mirroring `releaseEditorFocusAfterOpen`'s contract; menu-driven entry focuses the textarea at the mapped entry caret.

One behavior change this surfaced: the editor menu button was gated on share/save capability and disappeared entirely without the Android bridges (framework-free web runs included). The sheet now always carries at least the source-mode toggle, so the button always shows; capability rows still gate themselves.

## Verification

- Engine composition spec 8/8 (round-trip, boundary survival, canonicalization, raw-coordinate caret mapping with built-in counterfactual, syntax-boundary snapping); full Muya gate **1538/1538**.
- App unit **655/655**: controller (12 cases: entry capture, null fallbacks, exit hand-back with the raw text riding along, last-known-caret fallback, document-start snap when the engine ladder is exhausted, unchanged exit, teardown race, reset), caret math round-trip over every offset, session override (incl. single-newline normalization), Android-back priority.
- E2E: 5 new source-mode cases (round-trip with single-step undo via the toolbar, autosave-while-active hitting durable storage, setting-driven entry without focus, menu-driven entry with focus, theme-variable colors) plus the affected editor suites — **75/75** across toolbar/search/outline/modal-accessibility/android-document-actions/source-mode.
- **On-emulator walkthrough (Google AVD, API 35)**: share-intent injected document → menu shows the toggle → mode entered with canonical table serialization visible → typed text shows "Autosaving locally" while active → exit renders the new paragraph → a single toolbar undo removes the whole session and restores the entry caret. Dark theme verified throughout.


Renakoni:
Round 1 addressed in 436f929. The finding is real, and it stings precisely because my own audit had already pinned the ground truth it violates: the composition spec proved the exit hand-back re-serializes to CANONICAL markdown, yet the exit caret's `{line, ch}` was still computed against the RAW textarea text and handed to `setCursorByOffset`, which injects its sentinels into `getMarkdown()` — the canonical text. Two different serializations, one pair of coordinates: anywhere canonicalization changes line content (padded table columns) or line counts, the caret landed away from where the user left it.

**The fix moves the mapping into the engine, on principle rather than by patching offsets**: `setCursorByOffset(cursor, sourceMarkdown?)` now injects its position sentinels into the serialization the coordinates are actually expressed against — the raw textarea text — and resolves the located blocks in the clean rebuild as before. This is correct because both serializations parse to the SAME state (canonicalization is serializer-side only), so a block path + content offset found in the raw-sentinel parse is valid in the canonical tree; the sentinel mechanism locates the caret by block + content offset, not by string position, which is exactly what makes it immune to serialization drift. The source-mode exit passes the raw text; all other callers are unchanged (the parameter is optional and defaults to the old behavior).

**The requested hand-off regression is in the engine spec, with the counterfactual built into the test itself**: the same raw coordinates (caret after the `d` cell of an unpadded table) are resolved twice — WITH the source text, `getCursorOffset()` confirms the caret sits right after `d` in the canonical document; WITHOUT it (the old behavior), the test asserts the resolution provably fails or lands away from `d`. The discriminating power is therefore executed on every run, not claimed — the lesson from the #179 fixture inversion, applied.

**On merge order** (from the review discussion): #182 is stacked on #181 — its history contains #181's commit. Merge #181 first; #182's diff then reduces to its own commits. Merging #182 first would fold #181's changes in with it (and leave #181 as an empty PR under squash).

Verification: engine composition spec 7/7, full Muya gate 1537/1537, app unit 654/654 (controller assertions now pin the raw text riding along), source-mode E2E 5/5, typecheck clean.

Renakoni:
Round 2 addressed in 2bb8033 — all three items.

**P1: syntax-boundary caret loss — confirmed and fixed at both layers, with an explicit contract.** The repro is exactly right: End on a table row parks the caret after the closing pipe, the sentinel injected at that precise position breaks the row's syntax, the degraded parse cannot locate the caret, and the old code shrugged with a warning. Per your option 1, the position now maps to the nearest editable offset — but without syntax heuristics (parser emulation is how the previous generation of this feature died): the ENGINE walks the requested position backward one character at a time, each candidate validated by the parse itself, until one resolves or the bounded ladder (32 steps, clearing a four-column delimiter row) runs out. Contract: **the caret restores to the nearest editable position at or before where the user left it.** A caret inside a pure-syntax delimiter row walks across the line boundary into the header row and still resolves. Per your option 2's spirit, the APP adds a deterministic last resort: if even the ladder fails, the caret snaps to the document start and the warning reports the snap — never a silent loss.

The requested regression covers your exact repro — `| c | d |` with the caret at the row end (after the closing pipe), asserted to resolve into the last cell of the canonical row — plus the delimiter-row walk, in the engine's hand-off spec where the real Muya parse runs. The controller test pins the document-start fallback call. (Note on the delimiter case: the RESTORE succeeds and is asserted; `getCursorOffset`'s read-back is nullable for some landing blocks — its own documented behavior — so the read-back is asserted only when available.)

**Merge conflict — resolved by rebasing the stack.** #181 was squash-merged as 2bb84c4; the branch carried its own copy of that commit, which is what made `settingsContent.test.ts` conflict. Rebased onto main (git dropped the duplicate automatically, the two #182 commits replayed cleanly), pushed with `--force-with-lease`. GitHub now reports **MERGEABLE**, and all gates were re-run on the rebased head.

**P3: PR body — synced for real this time**: engine spec 8/8, Muya gate 1538/1538, app unit 655/655, and the hardened caret contract is now part of the design section rather than trailing in comments.

One environment note: the E2E suite runs via `pnpm exec playwright test` after `pnpm exec playwright install chromium` — the binary is a devDependency, which is why an isolated checkout without the install step reports `playwright not recognized`. On this head: source-mode E2E 5/5.

Verification on the rebased head: engine composition spec 8/8, full Muya gate 1538/1538, app unit 655/655, source-mode E2E 5/5, typecheck clean.

PR #186, opened 2026-08-06
Fix CJK bold rendering, loosen the feature list, name the theme count

Renakoni:
Three README improvements across all ten language versions, from maintainer review of the rendered pages.

## 1. Simplified Chinese bold was never rendering

Every bold lead-in on the zh-CN feature bullets displayed literal `**` (screenshot in review chat). Root cause is CommonMark's emphasis delimiter rule, not a typo: in `**绝不丢失你的内容。**自动保存` the closing `**` is preceded by punctuation (。) and followed by a non-punctuation CJK character with no space, which makes the run left-flanking and therefore **invalid as a closer** — the whole span stays literal. Chinese runs text flush together, so the usual "period inside the bold" English habit breaks rendering.

Fix: the full stop moves outside the closing marker (`**绝不丢失你的内容**。自动保存`) on all 10 affected bullets. Verified both directions against marked (same CommonMark delimiter rules GitHub uses): the old pattern produces no `<strong>`, the new one does. Traditional Chinese had dodged the bug with a space after `**` — it adopts the same punctuation placement and drops the workaround space, so both Chinese READMEs read identically tight.

## 2. The headline feature list breathes

The "轻量的 Markdown 编辑器，一字不丢" block (and its nine siblings) was a tight list. Blank lines between bullets make it render as a loose list with paragraph spacing, matching the airier second block.

## 3. "30+ themes" instead of "custom themes"

The theme catalog ships 32 themes; every language's themes bullet now says so ("30+", "über 30", "más de 30", "30 種類以上", "30가지가 넘는", …) instead of the vague "custom themes".

## Not changed: the tagline

The review question was whether "Markdown, quietly yours." should become "Markdown, quietly, yours". Kept as-is deliberately: "quietly yours" is a wordplay on the letter-closing formula "yours truly" — *quietly* modifies *yours* as one phrase, reading as "yours, in the quiet/private way", which is exactly the product's privacy pitch. Inserting a comma severs the phrase into two disconnected words and reads as a punctuation error to native ears.

## 4. Root directory cleanup

The nine translated READMEs move to `docs/readme/` — the repository root listed ten README files, more than half its entries. Only the English `README.md` stays at the root. All relative links (language switcher, screenshots, logo, RELEASING, both LICENSEs) are rewritten for the new depth and validated by a link checker across all ten files: zero broken targets.


Renakoni:
Second commit (008c344) folds in the root-directory cleanup from review: the nine translated READMEs move to `docs/readme/` (git renames, history preserved), leaving only the English README at the root. Every relative reference was rewritten for the new depth — the language switcher (English climbs to `../../README.md`, sibling translations stay same-directory), screenshots (`../screenshots/`), the logo/RELEASING links (`../assets/`, `../RELEASING.md`), and both LICENSE links (`../../`). A link checker validated every relative `href`/`src`/markdown link across all ten files: **zero broken targets** (it caught two real regression classes during the rewrite — the `docs/`-prefixed asset paths and an over-eager rewrite of sibling links — both fixed before commit).

Issue #158, opened 2026-07-16
Complete the low-API and PDF accessibility matrix


Renakoni:
**Verified so far:** long documents, local images, math/code, RTL PDF on an API 34 device; API 35 AVD; the v0.1.0 signed artifact passed install, same-key upgrade, cold launch, and draft retention.

**Still open:** any API 24-33 runtime verification; a full signed-artifact SAF/PDF/TalkBack rerun; low-memory devices. Chromium-printed PDFs currently have no structure tags — tagged PDF is only on the table if an alternative export engine is ever chosen (that trade-off should be recorded as a product decision).

Compatibility reports from real devices — especially older API levels — are welcome on this issue.

Renakoni:
## API 24 runtime verification (emulator)

First entry for the API 24-33 gap, on a fresh AVD (`android-24;google_apis;x86_64`, factory WebView Chrome 69):

1. **Instant native crash before splash** — `android:inset="15%"` in `ic_splash_logo.xml` is API 26+; resource resolution throws `Resources$NotFoundException` in `MainActivity.onCreate` on Android 7.0/7.1. Fixed in #190.
2. **Permanent blank screen after the splash fix** — the bundle targets Chrome 111 (Vite `baseline-widely-available`); WebView 69 dies at JS parse time and Capacitor silently loads the broken bundle anyway. #190 sets `minWebViewVersion: 111` plus a static error page, so out-of-date WebViews now get "update Android System WebView" guidance instead of a blank screen.

**Support floor made explicit by #190:** minSdk 24 *and* System WebView ≥ 111 (shipped March 2023; any device with Play services can reach it).

**Emulator boundary:** emulator images only ever carry the factory WebView — no image can reproduce the realistic low-API configuration (old OS + Play-updated WebView). On the API 24 AVD the app is now correctly gated, so SAF/PDF/editor smoke cannot run there. Functional verification of API 24-33 therefore stays dependent on real-device reports, which this issue already solicits.

Startup/low-memory measurements (#156) run separately on an API 35 AVD with constrained specs.

Renakoni:
## API 24 functional verification — completed in-emulator after all

[...]

PR #190, opened 2026-08-06
Fix API 24 launch: splash crash and old-WebView error page

Renakoni:
Part of #158 — the first runtime verification on API 24 (emulator, image `android-24;google_apis;x86_64`) found two launch blockers; this PR fixes both, and review round 1 hardened the WebView gate against a third gap.

## 1. Native crash before splash (API 24-25)

`drawable/ic_splash_logo.xml` used `android:inset="15%"` — fractional insets are an API 26+ feature, so resource resolution throws `Resources$NotFoundException` in `MainActivity.onCreate` and the app crashes instantly on Android 7.0/7.1, despite `minSdkVersion 24`.

**Fix:** the original fractional version moves to `drawable-v26/`; the base variant uses a fixed `18dp` inset. API 26+ keeps the exact same drawable as before, so no visual change on any currently working device.

## 2. Permanent blank screen on old System WebViews

After the splash fix the app reached the WebView and died at JS parse time (`Uncaught SyntaxError`): the bundle targets Vite's `baseline-widely-available` (**Chrome 111**), while the emulator's factory WebView is Chrome 69. Capacitor's default without `server.errorPath` is to log one line and load the broken bundle anyway — the user sees a blank screen forever.

**Fix:** `android.minWebViewVersion: 111` (matching the build target) plus `public/webview-error.html`, a plain HTML/CSS page (no JS, renders fine on WebView 53+).

## 3. Huawei WebView bypass (review round 1)

Capacitor judges `com.huawei.webview` against a separate config (`minHuaweiWebViewVersion`, default floor 10) using Huawei's own version scheme, so on Android 8+ Huawei devices an old Chromium core could pass the gate and still die parsing the bundle. Rather than guessing a Huawei-version-to-Chromium mapping (unreliable, rots), `index.html` now carries an ES5 inline check that reads the **real Chromium core version from the user agent** (`Chrome/NNN`, reported by every Chromium-based WebView including Huawei and Amazon) and redirects to the same error page below 111. The error page copy is provider-neutral accordingly.

Support floor after this PR: **minSdk 24 + any WebView with a Chromium 111+ core**. Note the ceiling fact: the last Chrome that installs on Android 7 is **119**, so the current target keeps an 8-version margin — raising the Vite target past 119 would kill API 24-25 support entirely.

## Verification (API 24 emulator)

- Factory WebView (Chrome 69): splash fix verified (no crash), Capacitor gate loads the error page (`Handling local request: .../webview-error.html`), renders cleanly.
- Gate-bypass simulation: with Chrome 119 installed and the UA threshold temporarily raised to 200, Capacitor's gate passes (119 ≥ 111) and the **UA check alone** redirects to the error page — exactly the Huawei failure mode, proven working.
- Real support path: official Google-signed Chrome 119 (last Android 7 release) installed via `adb install`; the app then runs fully — editor rendering (headings/bold/italic/lists), autosave, SAF `ACTION_CREATE_DOCUMENT` save (file verified on disk), PDF export (valid `%PDF-1.4` + share sheet), and share-intent import all pass on Android 7.0.
- API 35 emulator: gate does not misfire; app loads normally.
- `pnpm typecheck`, `eslint capacitor.config.ts`, `pnpm build`, `cap sync`, `assembleDebug` all clean.

### DIFF
--- a/android/app/src/main/res/drawable-v26/ic_splash_logo.xml
+++ b/android/app/src/main/res/drawable-v26/ic_splash_logo.xml
@@ -2,15 +2,10 @@
 <!--
   Splash icon: the launcher mark flattened to a single muted grey (its alpha
   silhouette, tinted), so the launch reads as a quiet monochrome mark rather than
-  the full-colour app icon.
-
-  API 24/25 fallback: fractional insets ("15%") are an API 26 InsetDrawable
-  feature and CRASH resource resolution on older releases (the app never
-  reached onCreate past the splash on Android 7). The v26 variant keeps the
-  fraction; this base variant approximates it with a fixed dp inset.
+  the full-colour app icon. Inset to sit inside the splash icon's safe zone.
 -->
 <inset xmlns:android="http://schemas.android.com/apk/res/android"
-    android:inset="18dp">
+    android:inset="15%">
     <bitmap
         android:src="@mipmap/ic_launcher_foreground"
         android:tint="@color/splash_icon_tint"
--- a/capacitor.config.ts
+++ b/capacitor.config.ts
@@ -8,6 +8,10 @@ const config: CapacitorConfig = {
     // The bundle targets Vite's baseline-widely-available (Chrome 111);
     // older System WebViews fail at JS parse time, so gate them to the
     // static error page below instead of a blank screen (#158).
+    // This gate only understands standard WebView version schemes; Huawei
+    // WebView is versioned independently, so index.html additionally checks
+    // the real Chromium core version from the user agent and redirects to
+    // the same error page.
     minWebViewVersion: 111
   },
   server: {
--- a/capacitor.config.ts
+++ b/capacitor.config.ts
@@ -3,7 +3,16 @@ import type { CapacitorConfig } from '@capacitor/cli';
 const config: CapacitorConfig = {
   appId: 'io.github.renakoni.marktextandroid',
   appName: 'MarkText',
-  webDir: 'dist'
+  webDir: 'dist',
+  android: {
+    // The bundle targets Vite's baseline-widely-available (Chrome 111);
+    // older System WebViews fail at JS parse time, so gate them to the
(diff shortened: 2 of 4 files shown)


Renakoni:
Review response — all three findings addressed at `ae1b413`.

**P1 (Huawei bypass): confirmed and fixed, but not via `minHuaweiWebViewVersion`.** Verified against the same Capacitor 8.4.1 source; one scoping note: the `getCurrentWebViewPackage()` branch only runs on Android 8+, so the gap affects Android 8+ Huawei-WebView devices specifically (API 24-25 use the per-package fallback and were never exposed). Instead of maintaining a Huawei-version-to-Chromium mapping — which has no reliable public basis and would rot as Huawei ships independently — `index.html` now carries an ES5 inline check that reads the real Chromium core version from the user agent (`Chrome/NNN`, reported truthfully by every Chromium-based WebView incl. Huawei/Amazon) and redirects to the same error page below the build target. That checks the thing that actually decides success (core syntax support) rather than any provider's version wrapper, and also covers any future 'Capacitor waves it through but the core is too old' case. `minWebViewVersion: 111` stays as the earlier, cleaner interception for standard WebViews.

Failure-path verified, not assumed: with Chrome 119 installed on the API 24 emulator and the UA threshold temporarily raised to 200, Capacitor's gate passes (119 ≥ 111) and the UA check alone lands on the error page — precisely the Huawei failure mode. Restored to 111, the app boots normally on the same device.

**P2 (provider-specific copy): reworded.** Now provider-neutral: 'MarkText could not start with this device's current WebView. Most often the WebView component is too old.' + app-store/system-updates guidance + 'Requires a WebView based on Chrome 111 or newer' (core-version phrasing, which stays accurate for Huawei's separate numbering).

**Non-blocking (`errorPath` also serves other main-frame errors): covered** — the new first line no longer asserts the cause; the version hint is framed as the most likely reason.

Issue #157, opened 2026-07-16
Minimal native instrumentation smoke tests


Renakoni:
Native coverage is JVM-only today. A minimal androidTest smoke should cover what JVM tests cannot: real SAF read/write/persist, incoming intents, and PDF completion — without duplicating existing JVM tests.

Renakoni:
PR #191 covers most of this issue with an androidTest smoke: real SAF read / atomic overwrite / fresh create through a live ContentResolver, an incoming-intent content URI read back through the real resolver, and PDF export through the platform print pipeline. All run without driving the system DocumentsUI (a tiny in-test ContentProvider backs the SAF URIs), and on a separate on-demand workflow so the slow emulator job stays off every PR.

**One item is intentionally left out of automation: SAF _persist_.** A faithful test would need a real persistable grant arriving from the system SAF flow (DocumentsUI) and then a process/emulator restart to prove the permission survives — exactly the flaky system-UI path this smoke avoids, and a self-granted permission across the test/app UID boundary proves nothing. Note the earlier on-device #158 run verified create/save + read-back, which is **not** the same as persistable-permission recovery.

So #191 is 'Part of #157', not a full close. Two reasonable ways to finish:
1. Narrow this issue's scope to exclude persist automation (justified above) and close it once #191 merges, tracking persist as a recorded manual check; or
2. Keep it open for a dedicated persist follow-up (manual: open via DocumentsUI → take persistable permission → restart → still read/write the same URI without re-picking).

Flagging so the completion claim matches the evidence — deciding which is a call for the maintainer.

Renakoni:
Closing per option 1 above, now that PR #191 is merged (6f4e007): the androidTest smoke covers real SAF read / atomic overwrite / fresh create via a live ContentResolver, incoming-intent content-URI read-back, and PDF export through the platform print pipeline, on the on-demand Android Connected Tests workflow.

Scope note: SAF persistable-permission recovery is deliberately excluded from automation (a faithful test needs a real DocumentsUI grant plus a restart — the flaky system-UI path this smoke avoids). It stays a recorded manual check: open via DocumentsUI, take the persistable grant, restart, confirm the same URI still reads/writes without re-picking.

PR #191, opened 2026-08-06
Add minimal native instrumentation smoke tests (#157)

Renakoni:
Part of #157.

Native coverage was JVM-only, which runs against fakes and cannot reach the ContentResolver-, intent-, and print-coupled paths. This adds a **minimal androidTest smoke** for those — no duplication of the existing JVM tests.

## What it covers (6 tests)

- **`SafDocumentIoInstrumentationTest`** — real SAF I/O through a live `ContentResolver`: read an existing document, atomically overwrite one (the full read-backup → truncate → write → fsync → length-verify path against a real content descriptor), and write a freshly created one (unique name + delete + absent-assertion, so it is a genuine create rather than an accidental overwrite on a reused emulator).
- **`IntentUriResolverReadInstrumentationTest`** — the seam between intent parsing and real I/O: an `ACTION_VIEW` / `ACTION_SEND` `Intent` is classified by `IncomingIntentParser` and the resulting content URI is **read back through the actual resolver**. It is deliberately **not** end-to-end dispatch — it does not launch `MainActivity`, pass the manifest filters, or exercise the plugin/JS path, so a broken filter or dropped grant would not be caught here (that is #158's on-device coverage). What it adds over the JVM `IncomingIntentParserTest` is that the parsed URI points at readable content.
- **`PdfExportInstrumentationTest`** — `PdfExporter.export()` completes through the platform print pipeline (WebView → `PrintDocumentAdapter` → `PdfDocument`) and writes a valid `%PDF-` file. Unreachable from JVM.

## Why this is `Part of`, not `Closes` #157

#157 lists **SAF read / write / persist**. This PR does **not** cover *persist*, and — correcting an earlier version of this description — #158 does **not** cover it either: #158 verified `ACTION_CREATE_DOCUMENT` save + read-back, which proves create/save but **not** `takePersistableUriPermission` surviving a process/emulator restart. A faithful persist test needs a real grant from the system SAF flow (DocumentsUI) plus a restart — the flaky system-UI path this design avoids — and the test provider is an unprotected `exported=true` provider, so it cannot stand in for a real grant.

Persist is therefore **out of automation scope** here and left as a decision on #157 (narrow the issue's scope with persist as a recorded manual check, or keep it open as a persist follow-up). Because of that this PR is **Part of #157** and does not auto-close it. See the comment on #157.

## How the SAF cases stay stable

The picker is the system DocumentsUI (a separate process, layout varies by API level), so `TestDocumentProvider` (a tiny in-test `ContentProvider` mapping a content URI to a real cache file) backs the URIs and everything is seeded/verified through the resolver. Test and provider run under different UIDs (app is `.debug`, test APK `.debug.test`); the provider is registered only in the androidTest manifest (`exported=true` so the app process can reach it) and never ships.

## CI

A separate **`android-connected.yml`** workflow runs these (`workflow_dispatch` + `android/**`), scoped to `:app:connectedDebugAndroidTest` so it does not build the unrelated `capacitor-cordova-android-plugins` androidTest variant. All 7 checks pass on a real API 35 emulator.

### DIFF
--- a/.github/workflows/android-connected.yml
+++ b/.github/workflows/android-connected.yml
@@ -1,0 +1,82 @@
+name: Android Connected Tests
+
+# Instrumentation (androidTest) smoke that needs a real device: SAF read/write
+# through a live ContentResolver, incoming-intent content URIs, and PDF export
+# through the platform print pipeline. The emulator job is slow (~8-10 min) and
+# occasionally flaky, so it runs on demand and only when native code changes —
+# not on every PR. See issue #157.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - "android/**"
+      - ".github/workflows/android-connected.yml"
+  push:
+    branches:
+      - main
+    paths:
+      - "android/**"
+      - ".github/workflows/android-connected.yml"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  connected-tests:
+    name: Instrumentation smoke
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/actions/setup
+
+      - name: Install JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: "22"
+          cache: gradle
+
+      # The debug APK bundles the built web assets, so produce and sync them
+      # before assembling the app under test.
+      - name: Build web shell
+        run: pnpm build
+
+      - name: Sync Capacitor Android assets
+        run: pnpm exec cap sync android
+
+      # Hardware acceleration for the emulator on the Linux runner.
+      - name: Enable KVM
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+
+      - name: Run instrumentation smoke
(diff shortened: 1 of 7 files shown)


Renakoni:
You're right on all three — and the reason they were missing is important: those fixes **were** committed and pushed last round, but a branch rewrite (a rebase/"Update branch" that produced `a106ce9`) dropped that commit, so the head you reviewed had reverted to the pre-fix state. Re-applied now at `5d33922`:

- **P3 (fresh-create determinism)** — `writesFreshlyCreatedDocument` now uses a unique name (`fresh-<nanoTime>.md`), deletes first, and asserts the document is absent before writing; `TestDocumentProvider.delete` really deletes. So it's a genuine create even on a reused emulator.
- **P2 (intent scope)** — renamed `IncomingIntentContentUriInstrumentationTest` → `IntentUriResolverReadInstrumentationTest` and rewrote the javadoc to say explicitly it covers the parser→resolver-read seam, **not** end-to-end dispatch (no MainActivity/manifest-filter/plugin-JS path — a broken filter or dropped grant would not be caught here; #158 covers that on device).
- **P2 (persist / completion claim)** — PR body changed from `Closes #157` to **`Part of #157`**, and the incorrect "#158 covers persist" claim is removed: #158 verified create/save + read-back, not `takePersistableUriPermission` surviving a restart. Persist is stated as out-of-automation-scope, with the narrow-scope-vs-follow-up decision left on #157 (comment already there).

On the **Kotlin duplicate-class** point from the previous round: I did **not** re-add the global `resolutionStrategy` — your original position (scope to `:app:`) is the right one and is sufficient. The current head passes the connected job on `:app:connectedDebugAndroidTest` alone (the `a106ce9` run succeeded), and a `--dry-run` of that task on current main does not pull `:capacitor-cordova-android-plugins:checkDebugAndroidTestDuplicateClasses`. The earlier failure was on a pre-rebase base; it's moot now, so no global dependency override is introduced.

androidTest compiles clean; pushing `5d33922` re-triggers the connected workflow.

PR #219, opened 2026-09-19
Checkpoint reading positions before lifecycle shutdown

### LINKED ISSUE #216: [Bug] Checkpoint reading positions before Android lifecycle shutdown
## Problem

Reading-position capture currently happens only when leaving the editor or on a lifecycle event. Scrolling marks the session as touched but writes no checkpoint. Even when a pause/hidden event reaches JavaScript, capture waits for an asynchronous SHA-256 digest before writing localStorage. Removing the task or losing the WebView process can interrupt that final operation.

A browser probe confirmed that scrolling to the middle and waiting three seconds creates no record, and interrupting a lifecycle capture while its digest is pending leaves no record on reopen.

## Scope and intended behavior

Persist bounded checkpoints during foreground reading/editing, retain an immediate live capture for existing exit/lifecycle paths, and reuse the fingerprint of unchanged Markdown. Keep the current document identity, storage schema, strict content validation, manual resume card, and structural block/ratio anchor.

This issue does not add native background services, automatic scrolling on reopen, approximate recovery after external edits, source-text position tracking, or document-content recovery changes. Source-mode corruption is tracked separately.

## Implementation checklist

- [x] Save after a short scroll/edit quiet period, with a maximum wait during continuous activity.
- [x] Read the live content and anchor together at capture time; avoid serializing or hashing on every scroll event.
- [x] Reuse a completed fingerprint only for identical Markdown; invalidate/recompute it for changed content and reset it with the session.
- [x] Keep latest-capture-wins ordering, outgoing-document identity, and passive-revisit preservation.
- [x] Cancel scheduled work on document replacement/teardown; an explicit lifecycle capture supersedes its scheduled checkpoint.
- [x] Add deterministic scheduling/ordering tests and mobile browser coverage that survives reload without depending on an exit callback.
- [x] Run relevant unit/E2E tests, build, lint, and document the process-death validation limits.

## Acceptance criteria

A reader who scrolls and pauses has a stored position before leaving the application. Continuous activity cannot indefinitely postpone initiating a checkpoint. Ending the page without a final lifecycle callback preserves the last completed checkpoint. Immediate exit captures the latest position, and stale asynchronous writes cannot replace newer captures or cross document identities. Reading-position writes do not dirty or save document content.

Renakoni:
## Summary

Reading positions previously depended on leaving the editor or delivering a lifecycle callback, then waiting for an asynchronous full-document SHA-256 digest. A task removal or process shutdown could interrupt that last operation. The editor now saves foreground checkpoints after activity settles and during continuous activity, so a completed position exists before shutdown begins.

## Behavior

- Scroll/edit activity saves after 500 ms of quiet. Continuous activity initiates a checkpoint at most 2.5 seconds after the first unsaved event; the quiet deadline cannot indefinitely postpone it.
- Existing exit/lifecycle paths still capture the latest live viewport immediately and cancel the scheduled checkpoint.
- Once a fingerprint has completed, unchanged Markdown reuses it and writes a fresh anchor synchronously, without another WebCrypto suspension.
- While an unchanged-content fingerprint is still computing, newer captures share that digest and use their own latest anchor and capture time. A slow hash can complete even when continuous scrolling triggers more checkpoints.
- Passive revisits do not overwrite the previous record. Source-mode suspension follows the capture-surface guard from #218.

## Design and responsibility boundaries

The resume session controller owns the two bounded timers, one completed-capture cache, and one reference to pending fingerprint work. Scroll handlers only mark activity and schedule work; they do not read DOM geometry, serialize Markdown, or hash the document on every event. A checkpoint flushes pending editor operations, reads current Markdown, and then captures the current structural block/ratio anchor in the same JavaScript turn.

Fingerprint reuse requires exact equality with the cached or pending Markdown, including same-length edits. It reuses only identity metadata: the anchor, display text, and timestamp are captured anew. Cached synchronous writes and asynchronous hash completions share the existing per-document sequence tokens, so even an undo to cached content supersedes an older in-flight edited-content digest. Sharing the pending digest avoids repeatedly invalidating hashes that take longer than the checkpoint interval. Failed digests release the pending reference so a later capture can retry. Both references are session-local and reset with the editor; a late outgoing write cannot populate the next session's cache or clear its pending hash.

Document replacement and teardown cancel both timers. Explicit persistence also clears timers scheduled reentrantly by flushing pending editor operations. No new storage schema, dependency, setting, native service, or document-content save path is introduced. Resume-card presentation and exact-content validation remain unchanged.

## Scope and integration

This PR targets `main`, which includes the source-mode safety fix from merged #218. Its diff and sole fix target are #216. The main-branch integration preserves the reviewed checkpoint implementation and regression coverage.

The timing bounds govern checkpoint initiation, not guaranteed disk durability or hash-completion deadlines. Abrupt termination can still lose activity newer than the last completed checkpoint, and changed content must complete a new digest. This does not change document autosave or promise recovery of unsaved Markdown.

## Verification

- Four scheduling/cache regressions failed against the original controller before implementation.
- Web unit tests after self-review: 83 files / 721 tests passed, including quiet/max-wait scheduling, edit-only activity, timer cancellation, passive revisits, synchronous cached exits, changed-content hashing, cached-vs-in-flight ordering, slow shared hashes, hash-failure retry, reentrant flush notifications, and late outgoing-session completions. A new delayed-hash regression failed before the pending-work sharing fix.
- Resume, source-mode, and local-draft lifecycle mobile browser E2E after self-review: 37/37 passed. New cases cover foreground persistence/reload with exit handlers suppressed, continuous scrolling, a same-turn lifecycle write with WebCrypto deliberately blocked after warming the cache, a 3-second digest during continuous checkpoints, and the parent PR's pending-offer cancellation across a source-mode round trip.
- `pnpm build` (includes TypeScript typecheck), ESLint on all 259 tracked JavaScript/TypeScript/Vue files, changed-file ESLint, and `git diff --check` passed.
- `cap sync android` and Android `assembleDebug` passed.
- Dedicated API 35 emulator, installed debug APK: foreground checkpoint confirmed before leaving; real recent-task swipe confirmed task removal and restored `scrollTop=3000`; Android Home/resume in source mode preserved the WYSIWYG block; foreground `am force-stop` confirmed process termination, then cold restart restored `scrollTop=5000` from the completed checkpoint. The swipe left a cached process alive, so process termination was tested separately rather than inferred from task removal.

The Android checks used a fresh isolated AVD. They do not establish behavior on every OEM or guarantee a final write during abrupt termination; recovery is from the last completed checkpoint. The unfiltered workstation `pnpm lint` sees pre-existing untracked archive files, so tracked-file lint was used without changing the lint rules.

Closes #216


Renakoni:
@codex help me resolve the conflicts

Renakoni:
@codex help me resolve the conflicts

